PR #17, opened 2023-06-26
Prohibit updating spec.databaseInstance

kajinamit:
Updating spec.databaseInstance does not migrate existing data to a new instance and can break resource access unexpectedly. This prohibits such update.

### DIFF
--- a/api/v1beta1/placementapi_types.go
+++ b/api/v1beta1/placementapi_types.go
@@ -46,9 +46,9 @@ type PlacementAPISpec struct {
 	ServiceUser string `json:"serviceUser"`
 
 	// +kubebuilder:validation:Required
-	// MariaDB instance name
-	// Right now required by the maridb-operator to get the credentials from the instance to create the DB
-	// Might not be required in future
+	// MariaDB instance name.
+	// This can't be updated unless advanced mode is enabled, because updating this does not migrate
+	// data to a new database instance.
 	DatabaseInstance string `json:"databaseInstance"`
 
 	// +kubebuilder:validation:Optional
@@ -85,6 +85,11 @@ type PlacementAPISpec struct {
 	// Debug - enable debug for different deploy stages. If an init container is used, it runs and the
 	// actual action pod gets started with sleep infinity
 	Debug PlacementAPIDebug `json:"debug,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// AdvancedMode - allows some usage not fully supported by this operator.
+	AdvancedMode bool `json:"advancedMode"`
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=false
--- a/api/v1beta1/placementapi_webhook.go
+++ b/api/v1beta1/placementapi_webhook.go
@@ -97,16 +97,18 @@ func (r *PlacementAPI) ValidateUpdate(old runtime.Object) error {
 		return apierrors.NewInternalError(fmt.Errorf("unable to convert existing object"))
 	}
 
-	if r.Spec.DatabaseInstance != oldPlacement.Spec.DatabaseInstance {
-		return apierrors.NewForbidden(
-			schema.GroupResource{
-				Group:    GroupVersion.WithKind("PlacementAPI").Group,
-				Resource: GroupVersion.WithKind("PlacementAPI").Kind,
-			}, r.GetName(), field.Forbidden(
-				field.NewPath("spec").Child("databaseInstance"),
-				"Value is immutable",
-			),
-		)
+	if ! r.Spec.AdvancedMode {
+		if r.Spec.DatabaseInstance != oldPlacement.Spec.DatabaseInstance {
+			return apierrors.NewForbidden(
+				schema.GroupResource{
+					Group:    GroupVersion.WithKind("PlacementAPI").Group,
+					Resource: GroupVersion.WithKind("PlacementAPI").Kind,
+				}, r.GetName(), field.Forbidden(
+					field.NewPath("spec").Child("databaseInstance"),
+					"Value is immutable.",
+				),
+			)
+		}
 	}
 
 	return nil
--- a/api/v1beta1/placementapi_webhook.go
+++ b/api/v1beta1/placementapi_webhook.go
@@ -22,8 +22,14 @@ limitations under the License.
 package v1beta1
 
 import (
+	"fmt"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
+
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
@@ -86,7 +92,23 @@ func (r *PlacementAPI) ValidateCreate() error {
 func (r *PlacementAPI) ValidateUpdate(old runtime.Object) error {
 	placementapilog.Info("validate update", "name", r.Name)
 
-	// TODO(user): fill in your validation logic upon object update.
+	oldPlacement, ok := old.(*PlacementAPI)
+	if !ok || oldPlacement == nil {
+		return apierrors.NewInternalError(fmt.Errorf("unable to convert existing object"))
+	}
+
+	if r.Spec.DatabaseInstance != oldPlacement.Spec.DatabaseInstance {
+		return apierrors.NewForbidden(
+			schema.GroupResource{
+				Group:    GroupVersion.WithKind("PlacementAPI").Group,
+				Resource: GroupVersion.WithKind("PlacementAPI").Kind,
+			}, r.GetName(), field.Forbidden(
+				field.NewPath("spec").Child("databaseInstance"),
+				"Value is immutable",
+			),
+		)
+	}
+
 	return nil
 }
 
--- a/test/functional/placementapi_webhook_test.go
+++ b/test/functional/placementapi_webhook_test.go
@@ -43,7 +43,7 @@ var _ = Describe("Placement webhook", func() {
 		placementAPI.Delete()
 	})
 
-	When("databaseInstance is being updated", func() {
+	When("databaseInstance is being updated with advanced mode default", func() {
 		It("should be blocked by the webhook and fail", func() {
 			instance := placementAPI.Instance
 			_, err := controllerutil.CreateOrPatch(
@@ -53,6 +53,35 @@ var _ = Describe("Placement webhook", func() {
 				})
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).Should(ContainSubstring("spec.databaseInstance: Forbidden: Value is immutable"))
+		})
+	})
+
+	When("databaseInstance is being updated with advanced mode disabled", func() {
+		It("should be blocked by the webhook and fail", func() {
+			instance := placementAPI.Instance
+
+			_, err := controllerutil.CreateOrPatch(
+				th.Ctx, th.K8sClient, instance, func() error {
+					instance.Spec.AdvancedMode = false
+					instance.Spec.DatabaseInstance = "changed"
+					return nil
+				})
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).Should(ContainSubstring("spec.databaseInstance: Forbidden: Value is immutable"))
+		})
+	})
+
+	When("databaseInstance is being updated with advanced mode disabled", func() {
+		It("should be blocked by the webhook and fail", func() {
+			instance := placementAPI.Instance
+
+			_, err := controllerutil.CreateOrPatch(
+				th.Ctx, th.K8sClient, instance, func() error {
+					instance.Spec.AdvancedMode = true
+					instance.Spec.DatabaseInstance = "changed"
+					return nil
+				})
+			Expect(err).NotTo(HaveOccurred())
 		})
 	})
 })
--- a/test/functional/placementapi_webhook_test.go
+++ b/test/functional/placementapi_webhook_test.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package functional_test
+
+import (
+	"os"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+var _ = Describe("Placement webhook", func() {
+
+	var placementAPI TestPlacementAPI
+
+	BeforeEach(func() {
+		// lib-common uses OPERATOR_TEMPLATES env var to locate the "templates"
+		// directory of the operator. We need to set them othervise lib-common
+		// will fail to generate the ConfigMap as it does not find common.sh
+		err := os.Setenv("OPERATOR_TEMPLATES", "../../templates")
+		Expect(err).NotTo(HaveOccurred())
+
+		placementAPI = NewTestPlacementAPI(TestNamespace)
+		placementAPI.Create()
+	})
+
+	AfterEach(func() {
+		placementAPI.Delete()
+	})
+
+	When("databaseInstance is being updated", func() {
+		It("should be blocked by the webhook and fail", func() {
+			instance := placementAPI.Instance
+			_, err := controllerutil.CreateOrPatch(
+				th.Ctx, th.K8sClient, instance, func() error {
+					instance.Spec.DatabaseInstance = "changed"
+					return nil
+				})
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).Should(ContainSubstring("spec.databaseInstance: Forbidden: Value is immutable"))
+		})
+	})
+})
--- a/test/functional/suite_test.go
+++ b/test/functional/suite_test.go
@@ -18,8 +18,12 @@ package functional_test
 
 import (
 	"context"
+	"crypto/tls"
+	"fmt"
+	"net"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo/v2"
@@ -168,6 +172,17 @@ var _ = BeforeSuite(func() {
 		Expect(err).ToNot(HaveOccurred(), "failed to run manager")
 	}()
 
+	// wait for the webhook server to get ready
+	dialer := &net.Dialer{Timeout: time.Duration(10) * time.Second}
+	addrPort := fmt.Sprintf("%s:%d", webhookInstallOptions.LocalServingHost, webhookInstallOptions.LocalServingPort)
+	Eventually(func() error {
+		conn, err := tls.DialWithDialer(dialer, "tcp", addrPort, &tls.Config{InsecureSkipVerify: true})
+		if err != nil {
+			return err
+		}
+		conn.Close()
+		return nil
+	}).Should(Succeed())
 })
 
 var _ = AfterSuite(func() {
--- a/test/functional/suite_test.go
+++ b/test/functional/suite_test.go
@@ -91,6 +91,13 @@ var _ = BeforeSuite(func() {
 			routev1CRDs,
 		},
 		ErrorIfCRDPathMissing: true,
+		WebhookInstallOptions: envtest.WebhookInstallOptions{
+			Paths: []string{filepath.Join("..", "..", "config", "webhook")},
+			// NOTE(gibi): if localhost is resolved to ::1 (ipv6) then starting
+			// the webhook fails as it try to parse the address as ipv4 and
+			// failing on the colons in ::1
+			LocalServingHost: "127.0.0.1",
+		},
 	}
 
 	// cfg is defined in this file globally.
@@ -127,13 +134,26 @@ var _ = BeforeSuite(func() {
 	Expect(th).NotTo(BeNil())
 
 	// Start the controller-manager if goroutine
+	webhookInstallOptions := &testEnv.WebhookInstallOptions
 	k8sManager, err := ctrl.NewManager(cfg, ctrl.Options{
 		Scheme: scheme.Scheme,
+		// NOTE(gibi): disable metrics reporting in test to allow
+		// parallel test execution. Otherwise each instance would like to
+		// bind to the same port
+		MetricsBindAddress: "0",
+		Host:               webhookInstallOptions.LocalServingHost,
+		Port:               webhookInstallOptions.LocalServingPort,
+		CertDir:            webhookInstallOptions.LocalServingCertDir,
+		LeaderElection:     false,
 	})
 	Expect(err).ToNot(HaveOccurred())
 
 	kclient, err := kubernetes.NewForConfig(cfg)
 	Expect(err).ToNot(HaveOccurred(), "failed to create kclient")
+
+	err = (&placementv1beta1.PlacementAPI{}).SetupWebhookWithManager(k8sManager)
+	Expect(err).NotTo(HaveOccurred())
+
 	err = (&controllers.PlacementAPIReconciler{
 		Client:  k8sManager.GetClient(),
 		Scheme:  k8sManager.GetScheme(),
--- a/api/bases/placement.openstack.org_placementapis.yaml
+++ b/api/bases/placement.openstack.org_placementapis.yaml
@@ -48,6 +48,11 @@ spec:
           spec:
             description: PlacementAPISpec defines the desired state of PlacementAPI
             properties:
+              advancedMode:
+                default: false
+                description: AdvancedMode - allows some usage not fully supported
+                  by this operator.
+                type: boolean
               containerImage:
                 description: PlacementAPI Container Image URL (will be set to environmental
                   default if empty)
@@ -60,9 +65,9 @@ spec:
                   file.
                 type: string
               databaseInstance:
-                description: MariaDB instance name Right now required by the maridb-operator
-                  to get the credentials from the instance to create the DB Might
-                  not be required in future
+                description: MariaDB instance name. This can't be updated unless advanced
+                  mode is enabled, because updating this does not migrate data to
+                  a new database instance.
                 type: string
               databaseUser:
                 default: placement
--- a/config/crd/bases/placement.openstack.org_placementapis.yaml
+++ b/config/crd/bases/placement.openstack.org_placementapis.yaml
@@ -48,6 +48,11 @@ spec:
           spec:
             description: PlacementAPISpec defines the desired state of PlacementAPI
             properties:
+              advancedMode:
+                default: false
+                description: AdvancedMode - allows some usage not fully supported
+                  by this operator.
+                type: boolean
               containerImage:
                 description: PlacementAPI Container Image URL (will be set to environmental
                   default if empty)
@@ -60,9 +65,9 @@ spec:
                   file.
                 type: string
               databaseInstance:
-                description: MariaDB instance name Right now required by the maridb-operator
-                  to get the credentials from the instance to create the DB Might
-                  not be required in future
+                description: MariaDB instance name. This can't be updated unless advanced
+                  mode is enabled, because updating this does not migrate data to
+                  a new database instance.
                 type: string
               databaseUser:
                 default: placement


SeanMooney:
/hold i understand why you are doing this but i was literally talking about how we could move form a single shared db to per service dbs if we needed too for scaleing reasons by doing a db dump/restores and updating the db reference.

so before we block this i want to at least discuss if we should

stuggi:
Maybe we want to do something similar to what we did in the infra-operator to ipset where we added an immutable [1] parameter. if `true` on the old spec, the webhook [2] blocks updating the CR spec, except of the immutable parameter. This forces an additional step to allow changes and sends a  "be careful" signal. Depending of the service operator/CR this immutable parameter could default to true/false.

[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_types.go#L48
[2] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_webhook.go#L102

kajinamit:
> /hold i understand why you are doing this but i was literally talking about how we could move form a single shared db to per service dbs if we needed too for scaleing reasons by doing a db dump/restores and updating the db reference.
> 
> so before we block this i want to at least discuss if we should

I understand that possible scenario to migrate database but if we don't have specific timeline to implement that support we should forbid this update. With current implementations, in case a user updates databaseInstance then it switches the database without data migration and it does not remove the existing database in the old instance properly. I'm more concerned about the potential problems caused by unexpected changes.
Anyway I'll bring this to the sync discussionsa this week.

kajinamit:
> Maybe we want to do something similar to what we did in the infra-operator to ipset where we added an immutable [1] parameter. if `true` on the old spec, the webhook [2] blocks updating the CR spec, except of the immutable parameter. This forces an additional step to allow changes and sends a "be careful" signal. Depending of the service operator/CR this immutable parameter could default to true/false.
> 
> [1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_types.go#L48 [2] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_webhook.go#L102

That's an option, I think. One thing we probably have to consider here is that we have to control whether a specific key should be immutable or not. We can probably add a dict option like debug and use it.

api/v1beta1/placementapi_webhook.go
```diff
@@ -86,7 +92,25 @@ func (r *PlacementAPI) ValidateCreate() error {
 func (r *PlacementAPI) ValidateUpdate(old runtime.Object) error {
```
gibizer, line 92: please note that this needs to be called from openstack-operator to meaningfully reject the update. See https://github.com/openstack-k8s-operators/openstack-operator/pull/355

test/functional/placementapi_webhook_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	When("databaseInstance is being updated", func() {
+		It("should be blocked by the webhook and fail", func() {
+			instance := placementAPI.Instance
+			_, err := controllerutil.CreateOrPatch(
+				th.Ctx, th.K8sClient, instance, func() error {
+					instance.Spec.DatabaseInstance = "changed"
+					return nil
+				})
+			Expect(err).To(HaveOccurred())
```
gibizer, line 54: Please check that the correct error is occurred. As this test now will pass even if the k8s returns http 500 for example.

test/functional/suite_test.go
```diff
@@ -127,13 +134,26 @@ var _ = BeforeSuite(func() {
...
+		LeaderElection:     false,
 	})
 	Expect(err).ToNot(HaveOccurred())
 
 	kclient, err := kubernetes.NewForConfig(cfg)
 	Expect(err).ToNot(HaveOccurred(), "failed to create kclient")
+
+	err = (&placementv1beta1.PlacementAPI{}).SetupWebhookWithManager(k8sManager)
+	Expect(err).NotTo(HaveOccurred())
+
```
gibizer, line 160: Please add a check that the webhook is up before start executing tests. See https://github.com/openstack-k8s-operators/nova-operator/blob/fe606377229a26837eb263cf99d3e7eb4d3b449b/test/functional/suite_test.go#L235-L245

gibizer (changes requested):

SeanMooney:
for what it's worth i like martins suggestion.

we can make this immutable by default after the creation of the CR.
timeline for supporting the data migration is probably not until next year.

I think we likely could support db instance migration before GA but we won't be prioritising this until we start looking at things like updates/upgrades in Q1 or later I think.

we would probably couple this with backup/restore, edge or disaster recovery day two operations.

gibizer:
> > Maybe we want to do something similar to what we did in the infra-operator to ipset where we added an immutable [1] parameter. if `true` on the old spec, the webhook [2] blocks updating the CR spec, except of the immutable parameter. This forces an additional step to allow changes and sends a "be careful" signal. Depending of the service operator/CR this immutable parameter could default to true/false.
> > [1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_types.go#L48 [2] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_webhook.go#L102
> 
> That's an option, I think. One thing we probably have to consider here is that we have to control whether a specific key should be immutable or not. We can probably add a dict option like debug and use it.

I have a bad feeling about the UX of having a separate map defining what field is settable and when. But I have no good suggestion how to force an extra round-trip for the user to make sure the DB change is intentional.

api/v1beta1/placementapi_webhook.go
kajinamit: Yes. I'm aware of that requirement.

test/functional/placementapi_webhook_test.go
kajinamit: Fixed in the latest version

test/functional/suite_test.go
kajinamit: Fixed in the latest version

api/v1beta1/placementapi_webhook.go
kajinamit: I'd work on that integration separately so I'll resolve this conversation now but please let me know if you want me to address anything early within this change.

api/v1beta1/placementapi_types.go
```diff
@@ -86,6 +86,11 @@ type PlacementAPISpec struct {
 	// actual action pod gets started with sleep infinity
 	Debug PlacementAPIDebug `json:"debug,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// PlacementAPIImmutable - make some specs immutable to prohibit update operations
+	// not fully supported by this operator.
+	Immutable PlacementAPIImmutable `json:"immutable,omitempty"`
```
gibizer: I have a bad feeling about the UX of having a separate struct / map defining what field is settable and when. But I have no good suggestion how to force an extra round-trip for the user to make sure the DB change is intentional.

Maybe(?) we can have a single "advanced-mode" flag for the whole CRD. That is defaulted to false, and prevents any dangerous operation. That would allow us not to duplicate fields directly.

api/v1beta1/placementapi_types.go
```diff
@@ -86,6 +86,11 @@ type PlacementAPISpec struct {
 	// actual action pod gets started with sleep infinity
 	Debug PlacementAPIDebug `json:"debug,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// PlacementAPIImmutable - make some specs immutable to prohibit update operations
```
gibizer: Please describe the logic at the DatabaseInstance field too.

api/v1beta1/placementapi_types.go
```diff
@@ -159,6 +164,14 @@ type PasswordSelector struct {
 	Service string `json:"service"`
 }
 
+// PlacementImmutable defines the observed state of PlacementImmutable
+type PlacementAPIImmutable struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=true
+	// Make databaseInstance immutable
+	DatabaseInstance bool `json:"databaseInstance"`
```
gibizer: This will not work well with webhooks and json ser/deserialization. The webhook ser/des step will default this to false (the golang default) when the field is not provided in the request. You need to make this a *bool to avoid this.

api/v1beta1/placementapi_webhook.go
```diff
@@ -86,7 +92,27 @@ func (r *PlacementAPI) ValidateCreate() error {
...
+
+	if r.Spec.DatabaseInstance != oldPlacement.Spec.DatabaseInstance {
+		if r.Spec.Immutable.DatabaseInstance {
+			return apierrors.NewForbidden(
+				schema.GroupResource{
+					Group:    GroupVersion.WithKind("PlacementAPI").Group,
+					Resource: GroupVersion.WithKind("PlacementAPI").Kind,
+				}, r.GetName(), &field.Error{
+					Type:     field.ErrorTypeForbidden,
+					Field:    "*",
```
gibizer: I think we should report the which field we are reporting the error on here. E.g. this should be "spec.databaseInstance" if the local webhook validated it, or "spec.placement.template.databaseInstance" if the openstack-operator validated it. You can pass a field.Path here to get the necessary base path of the field for the two different call path. See https://github.com/openstack-k8s-operators/nova-operator/pull/411/files#diff-0bb4bc5fee241020d547ea9eb775efb27b4d8082e09ad7d51a221cebb8cd6002R110 for example

test/functional/placementapi_webhook_test.go
```diff
@@ -0,0 +1,72 @@
...
+			instance := placementAPI.Instance
+			_, err := controllerutil.CreateOrPatch(
+				th.Ctx, th.K8sClient, instance, func() error {
+					instance.Spec.Immutable.DatabaseInstance = false
+					instance.Spec.DatabaseInstance = "changed"
+					return nil
+				})
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
```
gibizer, line 86: Please add a test case that shows that defaulting for the boolean field works as expected

gibizer (changes requested):

stuggi:
> > Maybe we want to do something similar to what we did in the infra-operator to ipset where we added an immutable [1] parameter. if `true` on the old spec, the webhook [2] blocks updating the CR spec, except of the immutable parameter. This forces an additional step to allow changes and sends a "be careful" signal. Depending of the service operator/CR this immutable parameter could default to true/false.
> > [1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_types.go#L48 [2] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_webhook.go#L102
> 
> That's an option, I think. One thing we probably have to consider here is that we have to control whether a specific key should be immutable or not. We can probably add a dict option like debug and use it.

do we have to be that specific on a per field via the CRD? why not just have a single `immutable` parameter. if that set all fields which should not be possible to change get blocked by the webhook. if for some reason one of those need to be changed the `immutable` param needs to set to false. in that case you could change any of them, but you have to do that additional step.

gibizer:
> > > Maybe we want to do something similar to what we did in the infra-operator to ipset where we added an immutable [1] parameter. if `true` on the old spec, the webhook [2] blocks updating the CR spec, except of the immutable parameter. This forces an additional step to allow changes and sends a "be careful" signal. Depending of the service operator/CR this immutable parameter could default to true/false.
> > > [1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_types.go#L48 [2] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_webhook.go#L102
> > 
> > 
> > That's an option, I think. One thing we probably have to consider here is that we have to control whether a specific key should be immutable or not. We can probably add a dict option like debug and use it.
> 
> do we have to be that specific on a per field via the CRD? why not just have a single `immutable` parameter. if that set all fields which should not be possible to change get blocked by the webhook. if for some reason one of those need to be changed the `immutable` param needs to set to false. in that case you could change any of them, but you have to do that additional step.

Yeah, that feels a bit easier UX. I said above:
> Maybe(?) we can have a single "advanced-mode" flag for the whole CRD. That is defaulted to false, and prevents any dangerous operation. That would allow us not to duplicate fields directly.

stuggi:
> > > > Maybe we want to do something similar to what we did in the infra-operator to ipset where we added an immutable [1] parameter. if `true` on the old spec, the webhook [2] blocks updating the CR spec, except of the immutable parameter. This forces an additional step to allow changes and sends a "be careful" signal. Depending of the service operator/CR this immutable parameter could default to true/false.
> > > > [1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_types.go#L48 [2] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_webhook.go#L102
> > > 
> > > 
> > > That's an option, I think. One thing we probably have to consider here is that we have to control whether a specific key should be immutable or not. We can probably add a dict option like debug and use it.
> > 
> > 
> > do we have to be that specific on a per field via the CRD? why not just have a single `immutable` parameter. if that set all fields which should not be possible to change get blocked by the webhook. if for some reason one of those need to be changed the `immutable` param needs to set to false. in that case you could change any of them, but you have to do that additional step.
> 
> Yeah, that feels a bit easier UX. I said above:
> 
> > Maybe(?) we can have a single "advanced-mode" flag for the whole CRD. That is defaulted to false, and prevents any dangerous operation. That would allow us not to duplicate fields directly.

right, could not connect the "advanced-mode" word with immutable on quick scanning your updates. we should just have the same param name across the operators. we used `Immutable`  https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_types.go#L46-L48  , but we only just introduced it, so can be changed if we agree on something else. I just realized, I missed a description there ...

kajinamit:
OK I agree that we can make it a single parameter. The reason why I preferred flags per parameters is that I wanted to make the flag name specific enough to help users understand the expectation.

Probably we want to name it `disableValidation` instead of `advancedMode`. The implication of advanced mode is quite vague and I'd prefer using more specific name, unless we have any other use cases we want to enable by that flag.

gibizer:
> > > > > Maybe we want to do something similar to what we did in the infra-operator to ipset where we added an immutable [1] parameter. if `true` on the old spec, the webhook [2] blocks updating the CR spec, except of the immutable parameter. This forces an additional step to allow changes and sends a "be careful" signal. Depending of the service operator/CR this immutable parameter could default to true/false.
> > > > > [1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_types.go#L48 [2] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_webhook.go#L102
> > > > 
> > > > 
> > > > That's an option, I think. One thing we probably have to consider here is that we have to control whether a specific key should be immutable or not. We can probably add a dict option like debug and use it.
> > > 
> > > 
> > > do we have to be that specific on a per field via the CRD? why not just have a single `immutable` parameter. if that set all fields which should not be possible to change get blocked by the webhook. if for some reason one of those need to be changed the `immutable` param needs to set to false. in that case you could change any of them, but you have to do that additional step.
> > 
> > 
> > Yeah, that feels a bit easier UX. I said above:
> > > Maybe(?) we can have a single "advanced-mode" flag for the whole CRD. That is defaulted to false, and prevents any dangerous operation. That would allow us not to duplicate fields directly.
> 
> right, could not connect the "advanced-mode" word with immutable on quick scanning your updates. we should just have the same param name across the operators. we used `Immutable` https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/network/v1beta1/ipset_types.go#L46-L48 , but we only just introduced it, so can be changed if we agree on something else. I just realized, I missed a description there ...

I guess the naming should reflect the effect of the flag. If we only control immutability with it then sure we should call it Immutable. But I'm wondering that there might be other situations where we want to disable complex validations. E.g. we have a validation webhook check in nova that you should not deploy metadata in cell0. So disabling such complex validation would not be an immutability change but more like an advanced operation that we only allow if you are super sure about it. (Note that the nova example might be a wrong one as I don't see why somebody want to ever deploy metadata on cell0, but similar complex rule might exists elsewhere).

gibizer:
> OK I agree that we can make it a single parameter. The reason why I preferred flags per parameters is that I wanted to make the flag name specific enough to help users understand the expectation.
> 
> Probably we want to name it `disableValidation` instead of `advancedMode`. The implication of advanced mode is quite vague and I'd prefer using more specific name, unless we have any other use cases we want to enable by that flag.

With `disableValidation` I have another problem (sorry it seems I have that guy today who does not like anything). We have validations we don't want to disable with this check. E.g. there is a cell name length validation that cannot be disabled as long cell names will always lead to CR creation error in nova-operator (as we are concatenating names). So in this case the vague naming might help to force us defining in the description of the field the exact meaning. I.e. which specific checks are disabled by that flag.

stuggi:
> > OK I agree that we can make it a single parameter. The reason why I preferred flags per parameters is that I wanted to make the flag name specific enough to help users understand the expectation.
> > Probably we want to name it `disableValidation` instead of `advancedMode`. The implication of advanced mode is quite vague and I'd prefer using more specific name, unless we have any other use cases we want to enable by that flag.
> 
> With `disableValidation` I have another problem (sorry it seems I have that guy today who does not like anything). We have validations we don't want to disable with this check. E.g. there is a cell name length validation that cannot be disabled as long cell names will always lead to CR creation error in nova-operator (as we are concatenating names). So in this case the vague naming might help to force us defining in the description of the field the exact meaning. I.e. which specific checks are disabled by that flag.

agreed, we should not skip all validations, correctness needs still to be checked.

SeanMooney:
i dont partcallay like disableValidation as we are not going to disabel all validation.
im -1 on having a flag to disable all validation. this is also not really a validation

a validation would simply be checking that the db instance exists and possibly that the table exists. 

maintenance mode or similar i think would be better capture that you are only performing a day two operation such as relocating the db. in mantance mode we could do something like scale down to 0 API instance until you set it to false so  the workflow would be:

1. enable maintance mode (scale down placement to prevent new requests)
2. dump the db
3. restore it to and different db server
4. update the reference to the new instance and unset maintance mode

in 1 we would scale down the API so that no new requests can happen 
2/3 would eventually be automated 
4 would restore api access with the container pointing at the new db.

any attempt to update the db outside of maintance mode would be blocked.

SeanMooney:
just to be clear if we add a maintenance mode option I'm ok with relaxing validation slightly during mantaince mode as long as we reseert it when exiting it.

gibizer:
I can compromise on "maintenanceMode" over "advancedMode" or "sureIKnowWhatI'mDoingMode" (OK, I'm kidding on the latter), but "maintenanceMode" still feels a bit too specific to one given process.

gibizer:
> just to be clear if we add a maintenance mode option I'm ok with relaxing validation slightly during mantaince mode as long as we reseert it when exiting it.

Ohh that is a very good point. We need to be careful if this mode allows persisting a value that otherwise would be rejected not because of the fact that it is a change but because of the specific new value itself, then we cannot re-enable the strict mode any more.

kajinamit:
I'd still disagree with the raw `maintenanceMode` because the aim of `maintenance` is not clear. We should probably name it `databaseMaintenanceMode` though that does not make much difference from the original proposal to have specific flag to make databaseInstance mutable ...

bshephar:
Just adding my 2 cents in here. But I feel like if we _need_ a maintenance mode, we might be making things a bit too complicated. Like you just discussed in the later comments, it introduces more complexity that we need to handle somewhere.

If we wanted some kind of migration facility, wouldn't it make more sense to provide some interface for that during the `reconcileUpdate` workflow? Since we expect to be handling schema updates and such during that update workflow, adding some migration workflow there would probably be easier? Rather than having some bespoke maintenance mode?

Also, does seperate DB's for each service mean we're going to run a mysql pod per service? That feels a bit heavy, and I'm not sure I understand what we're going to gain at the scale 90% of our users are at. Maybe in the like the top 5% of large deployments there would be some performance benefit.

stuggi:
> I'd still disagree with the raw `maintenanceMode` because the aim of `maintenance` is not clear. We should probably name it `databaseMaintenanceMode` though that does not make much difference from the original proposal to have specific flag to make databaseInstance mutable ...

the thing is will it only be the database, or is some other parameter coming next. Placement is simple and does not have a lot of params, but for other services there are probably more parameters which per default should not change after the initial create.

api/v1beta1/placementapi_types.go
kajinamit: Fixed in the latest version.

test/functional/placementapi_webhook_test.go
kajinamit: I've added the test case for default.

api/v1beta1/placementapi_webhook.go
kajinamit: Fixed in the latest version

api/v1beta1/placementapi_types.go
kajinamit: I've added the description of that field here: https://github.com/openstack-k8s-operators/placement-operator/pull/17/files#diff-e5c063113f25a841c51f8370d6711ece1751aa814b24f925f440881290cb6e59R170

Or do you intend to adding any additional description to this immutable key level ?

kajinamit:
Updated the PR to clear some of the earlier comments. We still have to agree with the overall design but I kind of feel we are unsure about the plan to support any day 2 operators. If we take the position to "avoid prohibiting anything we can later allow" then I'd probably abandon the whole my these validations works, just hoping users would not do anything unsupported.

bshephar:
> Updated the PR to clear some of the earlier comments. We still have to agree with the overall design but I kind of feel we are unsure about the plan to support any day 2 operators. If we take the position to "avoid prohibiting anything we can later allow" then I'd probably abandon the whole my these validations works, just hoping users would not do anything unsupported.

I'm just concerned that we will have users accidentally changing this and ending up in a mess. We need to consider the type of support cases that are going to come up, and what the actual recovery procedure will look like. My concern would be that if this particular issue come up, then it will automatically become an escalation and the recovery of it would be messy.

SeanMooney:
my original thought was in the future we would just automatically copy the db if you updated the DB instance for what its worth

I didn't think we would need to have any flag to enable that.

the reason i suggested maintenanceMode and not something like databaseMaintenanceMode
is I don't want to see a pattern of <funcaltiy>MantianceMode propagated in the operators

so if we had a flag at all I wanted to have one that was very generic.


I'm still not really sold on the idea that we need this to be imuatbale and would prefer to trust our users to understand that if the change the database reference it will print to a different db.

kajinamit:
> the reason i suggested maintenanceMode and not something like databaseMaintenanceMode is I don't want to see a pattern of MantianceMode propagated in the operators
> so if we had a flag at all I wanted to have one that was very generic.

I understand your points but this requires that we consider all of the `maintenance` tasks we support in all services, so that we understand what should be allowed in maintenance mode, which would too much expand scope of this work.

> would prefer to trust our users to understand that if the change the database reference it will print to a different db.

While maintaining TripleO, we've seen several unfortunate cases where deployments were messed up by wrong edit in  templates. Users might understand the impact of each edit(Note even some users had no idea about the changes they were making in their production, for real) but preventing any unintentional edit is merely impossible. Imagine the case where user do `oc edit openstackcontrolplane/openstack` and unintentionally modified `databasInstance: openstack` to `databaseInstance: openstac` while editing the spec by editor. I believe this can be quite possible scenario we can face.

kajinamit:
> Imagine the case where user do oc edit `openstackcontrolplane/openstack` and unintentionally modified `databasInstance: openstack` to `databaseInstance: openstac` while editing the spec by editor. 

Giving it another thought then this might not cause much serious result as it fails at an earlier point unless there is a DB instance with `openstack` deployed ... So we don't probably have to consider unintentional edit really.

Though I still feel guarding it is very beneficial for users as well as us. You know how many users changed AZ of existing compute nodes after they created instance on these hosts. Such "unsupported" operation can't be really blocked unless we prohibit these strictly at validation.

SeanMooney:
> changed AZ of existing compute nodes after they created instance on these hosts.

as an aside that specific example is a bug in the nova API i know how to fix that but we just have not yet
we missed blocking it in one specific edgecase. we do block that in most case except one specific case.

bogdando:
Adding immutability control flags look overcomplicating the API (CRD) design. There are better examples to follow, like immutability control over the [ReplicaSet specs](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates). The controller/validation logic should be taking care of it, not asked to be controlled by users.

Neither we need ``disableValidation`` flags in CRD, becase for that, the way to go is deploy w/o webhooks

stuggi:
> Adding immutability control flags look overcomplicating the API (CRD) design. There are better examples to follow, like immutability control over the [ReplicaSet specs](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates). The controller/validation logic should be taking care of it, not asked to be controlled by users.
> 
> Neither we need `disableValidation` flags in CRD, becase for that, the way to go is deploy w/o webhooks

not sure if I get this correct, but this won't work in our case. this would mean if there is a required change for any immutable parameter you'd have to recreate (delete/create) it, or do I get your point wrong?

kajinamit:
Giving it another thought after discussion with @stuggi and drafting some parameter descriptions, I now think the `advancedMode` is the most appropriate naming for the flag (among the ones proposed so far), because the flag basically allows usage not supported by the operator . Alternative option I can think of was `unsafe` or `unsupported` but these can imply different meanings.  Also I'd prefer advanced to maintenance because maintenance is too much vague.

api/v1beta1/placementapi_types.go
kajinamit: This is no longer applicable in the latest version.

api/v1beta1/placementapi_types.go
kajinamit: This is addressed in the latest version.

api/v1beta1/placementapi_webhook.go
```diff
@@ -86,7 +92,25 @@ func (r *PlacementAPI) ValidateCreate() error {
 func (r *PlacementAPI) ValidateUpdate(old runtime.Object) error {
 	placementapilog.Info("validate update", "name", r.Name)
 
-	// TODO(user): fill in your validation logic upon object update.
+	oldPlacement, ok := old.(*PlacementAPI)
+	if !ok || oldPlacement == nil {
+		return apierrors.NewInternalError(fmt.Errorf("unable to convert existing object"))
+	}
+
+	if ! r.Spec.AdvancedMode {
```
stuggi, line 100: do we want an explicite extra step to toggle the `advancedMode` ? then we should check this in the old object. if we are ok that updating `advancedMode` and other parameters in one go this is fine.

api/v1beta1/placementapi_types.go
```diff
@@ -46,9 +46,9 @@ type PlacementAPISpec struct {
 	ServiceUser string `json:"serviceUser"`
 
 	// +kubebuilder:validation:Required
-	// MariaDB instance name
-	// Right now required by the maridb-operator to get the credentials from the instance to create the DB
-	// Might not be required in future
+	// MariaDB instance name.
+	// This can't be updated unless advanced mode is enabled, because updating this does not migrate
+	// data to a new database instance.
```
gibizer, line 51: Will updating this to a new database instance at least ensures that the DB schema is applied to the new DB instance?

api/v1beta1/placementapi_webhook.go
```diff
@@ -86,7 +92,25 @@ func (r *PlacementAPI) ValidateCreate() error {
...
+				schema.GroupResource{
+					Group:    GroupVersion.WithKind("PlacementAPI").Group,
+					Resource: GroupVersion.WithKind("PlacementAPI").Kind,
+				}, r.GetName(), field.Forbidden(
+					field.NewPath("spec").Child("databaseInstance"),
+					"Value is immutable.",
+				),
+			)
+		}
+	}
```
gibizer, line 112: This logic needs to be in a function that is callable from openstack-operator validation webhook. The pattern is documented in https://github.com/openstack-k8s-operators/docs/blob/main/webhooks.md#reusing-webhook-logic

api/v1beta1/placementapi_webhook.go
```diff
@@ -86,7 +92,25 @@ func (r *PlacementAPI) ValidateCreate() error {
...
+
+	if ! r.Spec.AdvancedMode {
+		if r.Spec.DatabaseInstance != oldPlacement.Spec.DatabaseInstance {
+			return apierrors.NewForbidden(
+				schema.GroupResource{
+					Group:    GroupVersion.WithKind("PlacementAPI").Group,
+					Resource: GroupVersion.WithKind("PlacementAPI").Kind,
+				}, r.GetName(), field.Forbidden(
+					field.NewPath("spec").Child("databaseInstance"),
+					"Value is immutable.",
```
gibizer, line 108: I would mention advancedMode here.

api/v1beta1/placementapi_types.go
```diff
@@ -86,6 +86,11 @@ type PlacementAPISpec struct {
 	// actual action pod gets started with sleep infinity
 	Debug PlacementAPIDebug `json:"debug,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// AdvancedMode - allows some usage not fully supported by this operator.
```
gibizer, line 91: We need to be more explicit about documenting what is allowed.

bshephar: > We need to be more explicit about documenting what is allowed.

For me, this `advancedMode` terminology is too broad for this reason. We need to assume that at best users might read the documentation, but most likely they will just read the spec. And `advancedMode` doesn't tell me anything about what it does. Is this some kind of Placement advancedMode? Can I find out about it by reading the Placement docs? Does it allow the operator do advanced things? 

I personally would prefer to see something like feature sets, where we enable and disable specific features with clearly defined scopes. eg. `spec.enabledFeatures.databaseMigration`. This to me would be a more intuitive interface, and the expectations of such a setting are fairly obvious from the naming convention.

My other question would be, how common of an interface is this in other Kubernetes primitives or operators? Do we see boolean configurable "modes" elsewhere? Or is our implementation bespoke and therefore not quite as  intuitive to the average Kubernetes user? Keeping in mind that a large selling point for this solution is "You don't need to have TripleO specific knowledge / Heat specific knowledge / Red Hat specific knowledge. You just need to know kubernetes." I think from a design perspective, we should keep that fairly high on the priority list of considerations.

gibizer: > For me, this advancedMode terminology is too broad for this reason. We need to assume that at best users might read the documentation, but most likely they will just read the spec. And advancedMode doesn't tell me anything about what it does. Is this some kind of Placement advancedMode? Can I find out about it by reading the Placement docs? Does it allow the operator do advanced things?

It is the latter. :) 

On this point I'm OK to group the operator specific configuration if there are many (like advancedMode, preserveJobs) into a separate struct to express that they are distinct from the configuration that is for the deployed OpenStack service (like customeServiceConfig). 

> I personally would prefer to see something like feature sets, where we enable and disable specific features with clearly defined scopes. eg. spec.enabledFeatures.databaseMigration. This to me would be a more intuitive interface, and the expectations of such a setting are fairly obvious from the naming convention.

My problem with this that we might end up with a long list of flags if we start adding a configurable for each different "advanced feature". Like the next would be allowing changing the memcached instance under a flag. 

> My other question would be, how common of an interface is this in other Kubernetes primitives or operators? Do we see boolean configurable "modes" elsewhere? Or is our implementation bespoke and therefore not quite as intuitive to the average Kubernetes user? 

I'm not sure how we define 'boolean configurable "modes"' but it was fairly easy to find examples of boolean flags in Spec structs that's value affecting the behavior of the operator itself instead of affecting the behavior of the running service deployed by the operator:
* https://github.com/openstack-k8s-operators/rabbitmq-cluster-operator/blob/f8d9ac5a263c5d9c82dfeef8170eacf16c1cc5e1/api/v1beta1/rabbitmqcluster_types.go#L76-L80 
* JobSpec.Suspended https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#job-v1-batch 

> Keeping in mind that a large selling point for this solution is "You don't need to have TripleO specific knowledge / Heat specific knowledge / Red Hat specific knowledge. You just need to know kubernetes." I think from a design perspective, we should keep that fairly high on the priority list of considerations.

If we allow everything that k8s allows and ignoring what OpenStack supports from the set of operations k8s allows, then I think we will end up in too many dangerous possibilities. I think one of the value we can bring to the customers is codified knowledge about deploying OpenStack, encoded into the implementation of the operators. If we know that changing a DB server under a running OpenStack deployment is dangerous for the integrity of the deployment, then we should tell that to our users somehow. I also agree that we should not assume that they read the doc. So I would codify that knowledge into the operator.

bshephar: Hey @gibizer thanks for the response.

I think in both of those examples it's more clear what the setting is though. It's not a broadly scoped `advancedMode` is more my point there. I think we want to avoid a situation where people just think, "Advanced mode! I want to do advanced things! I'll set this to true."

The alternative of `databaseMigration` makes you think a little more. "Do I want to be doing a database migration? I don't think so. I'll leave it false."

It's not that we want to allow k8s things. We just want to make sure the interface is intuitive for people familiar with k8s. I'm not too sure there are examples of other operators implementing an `advancedMode` that allows an arbitrary list of operations is all. In this case, we're talking pretty specifically about a k8s operator feature where we would implement job logic to migrate databases around. The OpenStack service doesn't necessarily need to be aware or care, it just needs a new address for the DB service and a restart. So I think in this case, we're implementing an interface on a k8s CRD which seems more specific to a k8s implementation detail rather than OpenStack implementation detail.

But yeah, I understand your concern around an all encompassing list of possible things that can be done. I would guess at a max of 5 per service though which doesn't seem too bad.

In any case, I'm just a single voice. If we have a consensus on the approach proposed here, I'm happy to work with it. But I thought it would be worth at least raising my concern and suggestion for consideration before we merge it.

:)

Thanks again for the perspective and happy Monday.

gibizer: Hi @bshephar.
I have to admit your reasoning make sense and started to convince me to move away from advancedMode. If we can limit ourselves to a reasonable amount of  feature flags, and move them into a separate struct, like you proposed above with `spec.enabledFeatures` then I think I'm on board of this idea.

PR #48, opened 2023-08-04
Update placementapi to use service override

stuggi:
Removes creation of routes. Those get done in the openstack-operator. Via service overrides the openstack-operator customize the services for the endpoints to the desired state specified in the ctlplane CR using externalEndpoints or the service override in the service template.

Jira: [OSP-26690](https://issues.redhat.com//browse/OSP-26690)

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/313


api/v1beta1/placementapi_types.go
```diff
@@ -113,8 +114,14 @@ type PlacementAPISpec struct {
...
-	// ExternalEndpoints, expose a VIP using a pre-created IPAddressPool
-	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints,omitempty"`
+	// Override, provides the ability to override the generated manifest of several child resources.
+	Override APIOverrideSpec `json:"override,omitempty"`
+}
+
+// APIOverrideSpec to override the generated manifest of several child resources.
+type APIOverrideSpec struct {
+	// Override configuration for the Service created to serve traffic to the cluster.
+	Service []service.OverrideSpec `json:"service,omitempty"`
```
SeanMooney: i looked at https://github.com/openstack-k8s-operators/lib-common/pull/313 but didnt see where the service orride type is defined.

we probably want to be fairly strict in what can be overrinde if we expose this interface.
for example we likely do not want to allow the names to be modified but perhaps timeouts or service type might be ok

can you point me to the definition of this datastucture.
any change of the filed in this datastucture will eventually require the service operators to rev ther CRD version so  i would like to understand what exactly is in this before we add it to the CRD.

SeanMooney: is this it.

https://github.com/openstack-k8s-operators/lib-common/blob/539f4c7a835b9568087f688a1e3dd8ef91692b37/modules/common/service/types.go#L81-L100

if so i think that is overly broad as it seams to container the core service service spc which would mean any part fo the service could be overridden which is not a maintainable interface.

stuggi: yes, right now its label,annotations and the full spec. I kept it for now to be in sync with the rabbitmq operators, but we could e.g. limit the spec to Type, SessionAffinity,  loadBalancerSourceRanges, ExternalName, ExternalTrafficPolicy,  SessionAffinityConfig, IPFamilyPolicy, LoadBalancerClass and InternalTrafficPolicy

SeanMooney: i think i would prefer that. what I'm concerned about is an end user changing something that breaks our assumptions like the sector for the pods

stuggi:
closing in favor of #61

PR #53, opened 2023-08-15
Stop using KOLLA_CONFIG_FILE environment

kajinamit:


pkg/placement/dbsync.go
```diff
@@ -47,7 +47,6 @@ func DbSyncJob(
 	}
 
 	envVars := map[string]env.Setter{}
-	envVars["KOLLA_CONFIG_FILE"] = env.SetValue(KollaConfig)
```
gibizer, line 50: I think you can remove the global constant `KollaConfig` as well

gibizer (changes requested):

gibizer (approved):
Thanks. Looks good.

pkg/placement/dbsync.go
SeanMooney: done

SeanMooney (approved):

Issue #64, opened 2023-09-15
Align service config generation


gibizer:
Currently placement-operator generates the service config into a ConfigMap and uses an init container and ENV vars to add passwords to the config. To align with other operators we need to:
* remove the init container usage
* generate the full service config including passwords into a Secret

PR #92, opened 2023-10-13
[tlse] tls for PlacementAPI pod configuration

Deydra71:
Public/Internal service cert secrets and the CA bundle secret can be passed to configure httpd virtual hosts for tls termination. The certs get direct mounted to the appropriate place in etc/pki/tls/certs/%s.crt|key and a CA bundle to
/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem . Job deployments for bootstrap/cron get the CA bundle added if configured.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/428

Jira: [OSPRH-2368](https://issues.redhat.com/browse/OSPRH-2368)


templates/placementapi/config/placement-api-config.json
```diff
@@ -24,6 +24,27 @@
...
+        },
+        {
+            "source": "/var/lib/config-data/tls-certificates/tls.crt",
+            "dest": "/etc/pki/tls/certs/tls.crt",
+            "owner": "placement",
+            "perm": "0755",
+            "optional": true
+        },
+        {
+            "source": "/var/lib/config-data/ca-trust",
```
Deydra71: @stuggi 
Since we will be using one CA bundle file created by openstack-operator, we can specify the individual file instead  the general folder (?)

SeanMooney:
this PR should contian updates to the samples and kuttl and even tests to test it
before it can be merged

Deydra71:
> this PR should contian updates to the samples and kuttl and even tests to test it before it can be merged

It's in the draft state still ^^

SeanMooney:
yes i know just wanted to call that out
apparently the patches that added the service override functionality did not add that coverage so i don't want use  to continue to extend this functionality uhntil that gap is closed

templates/placementapi/config/placement-api-config.json
stuggi: yes lets do that

stuggi:
> yes i know just wanted to call that out apparently the patches that added the service override functionality did not add that coverage so i don't want use to continue to extend this functionality uhntil that gap is closed

I had added envtest coverage in serviceoverride PR https://github.com/openstack-k8s-operators/placement-operator/blob/main/tests/functional/placementapi_controller_test.go#L452 / https://github.com/openstack-k8s-operators/placement-operator/blob/main/tests/functional/placementapi_controller_test.go#L537

gibizer (commented):
The general direction looks good to me

Deydra71:
@gibizer  Could you please review, and give ok to test label if sufficient? I still need to make updates to kuttl and functional test, but otherwise the general TLS support is set from major part.

api/bases/placement.openstack.org_placementapis.yaml
```diff
@@ -365,6 +365,29 @@ spec:
...
+                    properties:
+                      caSecretName:
+                        type: string
+                    type: object
+                  service:
+                    description: Service contains server-specific TLS secret
+                    properties:
+                      disableNonTLSListeners:
+                        type: boolean
+                      secretName:
```
stuggi: I just realized that this is probably not enough. we create two services per api, public and internal to allow to use different service types (also LoadBalancer types if you do not use routes for public). With the current tls struct here, we only allow passing in one secret for the service operator.

stuggi: I mean this is not a blocker for this PR since its main focus is to get the cacert bundle in so that it can validate e.g.  tls connection to db/rabbit. the second step would be to enable the tls for placement itself. just something we have to enhance.

Deydra71: I see, updating the Service struct could bring more mess as it's already being used in other's implementations. I suggest keeping this PR focused on CA bundle support and then in next implement the support for the public and internal service type.

Deydra71:
On hold until the agreed changes of the TLS structure are implemented in the lib-common.

api/go.mod
```diff
@@ -67,3 +69,14 @@ require (
...
+replace ( //allow-merging
+	github.com/google/gnostic => github.com/google/gnostic v0.6.9
+	// pin to k8s 0.26.x for now
+	k8s.io/api => k8s.io/api v0.26.11
+	k8s.io/apimachinery => k8s.io/apimachinery v0.26.11
+	k8s.io/client-go => k8s.io/client-go v0.26.11
+	sigs.k8s.io/controller-runtime => sigs.k8s.io/controller-runtime v0.14.7
+)
+
+replace github.com/openstack-k8s-operators/lib-common/modules/common => github.com/stuggi/lib-common/modules/common v0.0.0-20231207114330-c01061824e7b
```
stuggi: can you update to the one which is from your lib-common PR and rerun `make manifests`

go.mod
```diff
@@ -90,3 +90,5 @@ replace github.com/openstack-k8s-operators/placement-operator/api => ./api
 // mschuppert: map to latest commit from release-4.13 tag
 // must consistent within modules and service operators
 replace github.com/openshift/api => github.com/openshift/api v0.0.0-20230414143018-3367bc7e6ac7 //allow-merging
+
+replace github.com/openstack-k8s-operators/lib-common/modules/common => github.com/stuggi/lib-common/modules/common v0.0.0-20231207114330-c01061824e7b
```
stuggi: same

pkg/placement/dbsync.go
```diff
@@ -44,6 +44,18 @@ func DbSyncJob(
...
 	envVars["KOLLA_BOOTSTRAP"] = env.SetValue("true")
 
+	// create Volume and VolumeMounts
+	volumes := getVolumes(instance.Name)
+	volumeMounts := getVolumeMounts()
+	initVolumeMounts := getInitVolumeMounts()
+
+	// add CA cert if defined
+	if instance.Spec.TLS.CaBundleSecretName != "" {
+		volumes = append(getVolumes(instance.Name), instance.Spec.TLS.CreateVolume())
```
stuggi: should be
`volumes = append(volumes, instance.Spec.TLS.CreateVolume())` instead of calling the func again.

same for the bellow lines

pkg/placement/deployment.go
```diff
@@ -74,12 +87,49 @@ func Deployment(
...
+	// create Volume and VolumeMounts
+	volumes := getVolumes(instance.Name)
+	volumeMounts := getVolumeMounts()
+	initVolumeMounts := getInitVolumeMounts()
+
+	// add CA cert if defined
+	if instance.Spec.TLS.CaBundleSecretName != "" {
+		volumes = append(getVolumes(instance.Name), instance.Spec.TLS.CreateVolume())
+		volumeMounts = append(getVolumeMounts(), instance.Spec.TLS.CreateVolumeMounts(nil)...)
+		initVolumeMounts = append(getInitVolumeMounts(), instance.Spec.TLS.CreateVolumeMounts(nil)...)
```
stuggi: same as mentioned above. no need to call the funcs again. just, append the tls stuff

pkg/placement/deployment.go
```diff
@@ -16,26 +16,39 @@ limitations under the License.
...
 func Deployment(
+	ctx context.Context,
+	helper *helper.Helper,
 	instance *placementv1.PlacementAPI,
 	configHash string,
 	labels map[string]string,
 	annotations map[string]string,
-) *appsv1.Deployment {
+) (*appsv1.Deployment, error) {
+	runAsUser := int64(0)
```
stuggi: I think this is not wanted. this would revert things from https://github.com/openstack-k8s-operators/placement-operator/commit/8fdff269484cd58ca9271f44b785a48647299f12

tests/functional/base_test.go
```diff
@@ -82,6 +85,15 @@ func CreateNames(placementAPIName types.NamespacedName) Names {
 		RoleBindingName: types.NamespacedName{
 			Namespace: placementAPIName.Namespace,
 			Name:      "placement-" + placementAPIName.Name + "-rolebinding"},
+		CaBundleSecretName: types.NamespacedName{
+			Namespace: placementAPIName.Namespace,
+			Name:      "combined-ca-bundle"},
```
stuggi: maybe rename the const and use them here as well and not just in the GetTLSPlacementAPISpec?

tests/functional/suite_test.go
```diff
@@ -112,6 +118,8 @@ var _ = BeforeSuite(func() {
 		},
 	}
 
+	logger = ctrl.Log.WithName("---Test---")
```
stuggi: already there in line 140

config/samples/placement_v1beta1_placementapi.yaml
```diff
@@ -14,6 +14,14 @@ spec:
     service: false
   preserveJobs: false
   replicas: 1
+  # tls:
```
SeanMooney: instead of commeting this out we should provide a separate tlse example file

pkg/placement/initcontainer.go
SeanMooney, line 1: we evenutally want ot remove all init container form this operator so can we avoid adding new usage of them?

SeanMooney: https://github.com/openstack-k8s-operators/placement-operator/issues/64

pkg/placement/volumes.go
```diff
@@ -94,7 +94,7 @@ func getVolumeMounts(serviceName string) []corev1.VolumeMount {
 		{
 			Name:      "config-data-merged",
 			MountPath: "/var/lib/kolla/config_files/config.json",
-			SubPath:   "placement-" + serviceName + "-config.json",
+			SubPath:   "placement-api-config.json",
```
stuggi: needs to be reverted as there are two kolla configs, one for the dbsync and one for the api

pkg/placement/dbsync.go
```diff
@@ -44,6 +44,18 @@ func DbSyncJob(
 	envVars["KOLLA_CONFIG_STRATEGY"] = env.SetValue("COPY_ALWAYS")
 	envVars["KOLLA_BOOTSTRAP"] = env.SetValue("true")
 
+	// create Volume and VolumeMounts
+	volumes := getVolumes(instance.Name)
+	volumeMounts := getVolumeMounts()
```
stuggi: should be `getVolumeMounts("dbsync")`

pkg/placement/deployment.go
```diff
@@ -74,12 +81,46 @@ func Deployment(
...
+		}
 	}
 
 	envVars := map[string]env.Setter{}
 	envVars["KOLLA_CONFIG_STRATEGY"] = env.SetValue("COPY_ALWAYS")
 	envVars["CONFIG_HASH"] = env.SetValue(configHash)
 
+	// create Volume and VolumeMounts
+	volumes := getVolumes(instance.Name)
+	volumeMounts := getVolumeMounts()
```
stuggi: should be `getVolumeMounts("api")`

controllers/placementapi_controller.go
```diff
@@ -318,8 +321,73 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	}
+)
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *PlacementAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// index passwordSecretField
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &placementv1.PlacementAPI{}, passwordSecretField, func(rawObj client.Object) []string {
+		// Extract the secret name from the spec, if one is provided
+		cr := rawObj.(*placementv1.PlacementAPI)
+		if cr.Spec.Secret == "" {
```
gibizer, line 350: I'm wondering if we can do some generic code that uses the constant field selectors (like passwordSecretField)  to access the field. So instead of having 4 blocks copy pasted here for the 4 secret, we could have a single loop that loops over all the field selectors. This can be considered in a follow up.

controllers/placementapi_controller.go
```diff
@@ -770,6 +876,38 @@ func (r *PlacementAPIReconciler) reconcileNormal(ctx context.Context, instance *
...
+		hash, ctrlResult, err := tls.ValidateCACertSecret(
+			ctx,
+			helper.GetClient(),
+			types.NamespacedName{
+				Name:      instance.Spec.TLS.CaBundleSecretName,
+				Namespace: instance.Namespace,
+			},
+		)
+		if err != nil {
+			return ctrlResult, err
```
gibizer, line 904: I think we should set the InputReadyCondition here to False with a message that point out which input is missing. Otherwise it is hard to tell from the CR status why the deplyoment is not progressing.

controllers/placementapi_controller.go
```diff
@@ -770,6 +876,38 @@ func (r *PlacementAPIReconciler) reconcileNormal(ctx context.Context, instance *
...
+
+		if hash != "" {
+			configMapVars[tls.CABundleKey] = env.SetValue(hash)
+		}
+	}
+
+	// Validate API service certs secrets
+	certsHash, ctrlResult, err := instance.Spec.TLS.API.ValidateCertSecrets(ctx, helper, instance.Namespace)
+	if err != nil {
+		return ctrlResult, err
```
gibizer, line 923: ditto.

controllers/placementapi_controller.go
```diff
@@ -851,8 +990,14 @@ func (r *PlacementAPIReconciler) reconcileNormal(ctx context.Context, instance *
...
 	// Define a new Deployment object
-	deplDef := placement.Deployment(instance, inputHash, serviceLabels, serviceAnnotations)
+	deplDef, err := placement.Deployment(ctx, helper, instance, inputHash, serviceLabels, serviceAnnotations)
 	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.DeploymentReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.DeploymentReadyErrorMessage,
+			err.Error()))
```
gibizer, line 1019: +1 thanks for propagating the error to the status.

templates/placementapi/config/httpd.conf
```diff
@@ -28,22 +28,47 @@ TransferLog /dev/stdout
 CustomLog /dev/stdout combined env=!forwarded
 CustomLog /dev/stdout proxy env=forwarded
 
+{{ range $endpt, $vhost := .VHosts }}
+# {{ $endpt }} vhost {{ $vhost.ServerName }} configuration
```
gibizer, line 32: nit: is this some debugging leftover?

gibizer (changes requested):
I only looked at the implementation I still have to look at the test code. I request changes in this PR only for the InputReady condition setting. My refactoring suggestions totally OK to be considered for a follow up.

controllers/placementapi_controller.go
stuggi: I previously had a quick look to use something like `reflect.Indirect(v).FieldByName(field)` but it did not work so we considered that a later optimization work to look into. agreed to check this in a follow up

controllers/placementapi_controller.go
stuggi: right, makes sense

templates/placementapi/config/httpd.conf
stuggi: this is intentional to have a comment in the httpd.conf for the virtual host. if it is considered to be useless we can remove it

controllers/placementapi_controller.go
stuggi: @Deydra71 check my keystone pr, there I had already added a TLSInputReady condition. lets also get this into the PR

templates/placementapi/config/httpd.conf
gibizer: Ahh, at first I thought that this is a comment in the template, but actually it is a comment template that is expaneded by the config generation to get a comment in the resulting config file. So this is good. Keep it. :)

controllers/placementapi_controller.go
gibizer: Thanks @Deydra71 for the fix.

tests/functional/base_test.go
```diff
@@ -92,6 +104,25 @@ func GetDefaultPlacementAPISpec() map[string]interface{} {
...
 
+func GetTLSPlacementAPISpec() map[string]interface{} {
+	return map[string]interface{}{
+		"databaseInstance": "openstack",
+		"replicas":         1,
+		"secret":           SecretName,
+		"tls": map[string]interface{}{
+			"api": map[string]interface{}{
+				"internal": map[string]interface{}{
+					"secretName": InternalCertSecretName,
```
gibizer: you can use `names.InternalCertSecretName` and the other refs from the Names struct here. There is always a global names variable with the naming constants grouped together (it is created in suite_test). If you use `names` here then you can inline the test global `InternalCertSecretName` into CreateNames(). I know that this sounds like meaningless busy work and for a true constant like how InternalCertSecretName is defined now it probably is. The reason we have the Names struct is to encapsulate generated names like Role and RoleBinding. (nova-operator has a bunch more examples) to avoid replicating the generation logic across multiple test case.

Anyhow I let you decide which way to go as it is not a big deal.

tests/functional/placementapi_controller_test.go
gibizer, line 1: Thanks you for adding env test coverage. I much appreciate the effort. Also the test looks good!

tests/kuttl/common/tls_certificates.yaml
```diff
@@ -0,0 +1,31 @@
+# Hardcoded certs secret, so kuttl doesn't require cert-manager at test runtime
```
gibizer, line 1: I agree that this is a good compromise to keep the test independent.

tests/kuttl/tests/placement_deploy_tls/03-assert.yaml
```diff
@@ -0,0 +1,239 @@
...
+    status: "True"
+    type: RoleReady
+  - message: ServiceAccount created
+    reason: Ready
+    status: "True"
+    type: ServiceAccountReady
+  - message: Service config create completed
+    reason: Ready
+    status: "True"
+    type: ServiceConfigReady
```
gibizer, line 85: the new TLSInputReady condition is missing from here

gibizer (approved):
This PR looks solid. The kuttl test needs a fix but other than that this is good to land from my perspective. Thanks for the effort!

tests/functional/base_test.go
Deydra71: I see, thanks for the explanation! Will edit it, to preserve same logic within tests.

controllers/placementapi_controller.go
SeanMooney: ok so i agree this can  be refactored to reduce the duplications

the lamda/funciton passed to IndexField  is just returning a StringSlice/charArray or nil if the filed is not set.

we know the mapphng between the string constants and the struct fields so we just need a pointer to member data for each and store a map of "string" -> "pointer to member data" which we can apply the the spec object to select the data 

if this was c++ that would be simple in golang I'm not sure that this easlisly do able so a slight different approach is needed 
```
field_indexers := map[string]interface{}{
    passwordSecretField: func(rawObj client.Object) []string {
		// Extract the secret name from the spec, if one is provided
		cr := rawObj.(*placementv1.PlacementAPI)
		if cr.Spec.Secret == "" {
			return nil
		}
		return []string{cr.Spec.Secret}
	},
	caBundleSecretNameField: func(rawObj client.Object) []string {
		// Extract the secret name from the spec, if one is provided
		cr := rawObj.(*placementv1.PlacementAPI)
		if cr.Spec.TLS.CaBundleSecretName == "" {
			return nil
		}
		return []string{cr.Spec.TLS.CaBundleSecretName}
	},
	tlsAPIInternalField: func(rawObj client.Object) []string {
		// Extract the secret name from the spec, if one is provided
		cr := rawObj.(*placementv1.PlacementAPI)
		if cr.Spec.TLS.API.Internal.SecretName == nil {
			return nil
		}
		return []string{*cr.Spec.TLS.API.Internal.SecretName}
	},
	tlsAPIPublicField: func(rawObj client.Object) []string {
		// Extract the secret name from the spec, if one is provided
		cr := rawObj.(*placementv1.PlacementAPI)
		if cr.Spec.TLS.API.Public.SecretName == nil {
			return nil
		}
		return []string{*cr.Spec.TLS.API.Public.SecretName}
	}
}

for field, indexer := range field_indexers{
    if err := mgr.GetFieldIndexer().IndexField(context.Background(), &placementv1.PlacementAPI{}, field, indexer
);err != nil {
		return err
	}

}
```

i think that is the same and will work but its not much simpler

the functor objects/lamdas are basically just c++ pointer to data members

SeanMooney: i have an example of this patteren here

https://github.com/openstack-k8s-operators/nova-operator/blob/8f204945b887a562229143f40901c4540925e9cd/controllers/common.go#L294-L341

tests/kuttl/common/tls_certificates.yaml
```diff
@@ -0,0 +1,31 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: combined-ca-bundle
+  labels:
+     service: placement
```
bogdando: indent

tests/kuttl/common/tls_certificates.yaml
```diff
@@ -0,0 +1,31 @@
...
...TjUrK09EN1I2MW5Gb1lBY0QKenpvUS91SW93NktjeGhwRWNQTDFxb3ZZUGxUYUJabEh3c2FpNE50VHA4aDA1RHVRSGZKOE9JNXFvMEl3UURBTwpCZ05WSFE4QkFmOEVCQU1DQXFRd0R3WURWUjBUQVFIL0JBVXdBd0VCL3pBZEJnTlZIUTRFRmdRVXE3TGtFSk1TCm1MOVpKWjBSOUluKzZkclhycEl3Q2dZSUtvWkl6ajBFQXdJRFJ3QXdSQUlnVlN1K00ydnZ3QlF3eTJHMVlhdkkKQld2RGtSNlRla0I5U0VqdzJIblRSMWtDSUZSNFNkWGFPQkFGWjVHa2RLWCtSY2IzaDFIZm52eFJEVW96bTl2agphenp3Ci0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0KIyByb290Y2EtcHVibGljCi0tLS0tQkVHSU4gQ0VSVElGSUNBVEUtLS0tLQpNSUlCZXpDQ0FTS2dBd0lCQWdJUU5IREdZc0JzNzk4aWJERDdxL28ybGpBS0JnZ3Foa2pPUFFRREFqQWVNUnd3CkdnWURWUVFERXhOeWIyOTBZMkV0YTNWMGRHd3RjSFZpYkdsak1CNFhEVEkwTURFeE5URXdNVFV6TmxvWERUTTAKTURFeE1qRXdNVFV6Tmxvd0hqRWNNQm9HQTFVRUF4TVRjbTl2ZEdOaExXdDFkSFJzTFhCMVlteHBZekJaTUJNRwpCeXFHU000OUFnRUdDQ3FHU000OUF3RUhBMElBQkQ3OGF2WHFocmhDNXc4czlXa2Q0SXBiZUV1MDNDUitYWFVkCmtEek9SeXhhOXdjY0lkRGl2YkdKakpGWlRUY1ZtYmpxMUJNWXNqcjEyVUlFNUVUM1ZscWpRakJBTUE0R0ExVWQKRHdFQi93UUVBd0lDcERBUEJnTlZIUk1CQWY4RUJUQURBUUgvTUIwR0ExVWREZ1FXQkJUS0ppeldVSjllVUtpMQpkczBscjZjNnNEN0VCREFLQmdncWhrak9QUVFEQWdOSEFEQkVBaUJJWndZcTYxQnFNSmFCNlVjRm9Sc3hlY3dICjV6L3pNT2RyT3llMG1OaThKZ0lnUUxCNHdES3JwZjl0WDJsb00rMHVUb3BBRFNZSW5yY2ZWdTRGQnVZVTNJZz0KLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo=
```
bogdando: indent

tests/kuttl/common/tls_certificates.yaml
```diff
@@ -0,0 +1,31 @@
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cert-internal-svc
+  labels:
+     service: placement
```
bogdando: indent

tests/kuttl/common/tls_certificates.yaml
```diff
@@ -0,0 +1,31 @@
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cert-public-svc
+  labels:
+     service: placement
```
bogdando: ditto

controllers/placementapi_controller.go
```diff
@@ -318,8 +322,73 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		if cr.Spec.Secret == "" {
+			return nil
+		}
+		return []string{cr.Spec.Secret}
+	}); err != nil {
+		return err
+	}
+
+	// index caBundleSecretNameField
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &placementv1.PlacementAPI{}, caBundleSecretNameField, func(rawObj client.Object) []string {
```
bogdando, line 359: could you please query this once (the whole spec) and store in a local var? Then you can fetch by needed fields

controllers/placementapi_controller.go
```diff
@@ -851,8 +1004,14 @@ func (r *PlacementAPIReconciler) reconcileNormal(ctx context.Context, instance *
 	//
 
 	// Define a new Deployment object
-	deplDef := placement.Deployment(instance, inputHash, serviceLabels, serviceAnnotations)
+	deplDef, err := placement.Deployment(ctx, helper, instance, inputHash, serviceLabels, serviceAnnotations)
```
bogdando, line 1012: ``deplDef, err =`` ?

controllers/placementapi_controller.go
```diff
@@ -338,6 +407,39 @@ func (r *PlacementAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+func (r *PlacementAPIReconciler) findObjectsForSrc(src client.Object) []reconcile.Request {
```
gibizer, line 415: This should be added as a Watch in via the NewControllerManagedBy(mgr) builder otherwise the TLS secrets are not watched.

```golang
		Watches(
			&source.Kind{Type: &corev1.Secret{}},
			handler.EnqueueRequestsFromMapFunc(r.findObjectsForSrc),
			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
		).
```

gibizer: Here is a commit with a test case that shows that the secret watching does not work without it https://github.com/gibizer/placement-operator/commit/f8ae8eea22a204157f84c779facda9e9d6d12533

gibizer (changes requested):

controllers/placementapi_controller.go
Deydra71: Not sure what you mean, can you please elaborate?

controllers/placementapi_controller.go
Deydra71: I'm unsure what you mean, could you please explain more?

controllers/placementapi_controller.go
Deydra71: Thanks a lot for raising this and the example, I added it to the other tests.

api/go.mod
```diff
@@ -67,3 +67,12 @@ require (
 // mschuppert: map to latest commit from release-4.13 tag
 // must consistent within modules and service operators
 replace github.com/openshift/api => github.com/openshift/api v0.0.0-20230414143018-3367bc7e6ac7 //allow-merging
+
+replace ( //allow-merging
+	github.com/google/gnostic => github.com/google/gnostic v0.6.9
+	// pin to k8s 0.26.x for now
```
gibizer: Hm, I'm wondering what pulling in newer k8s versions. Nothing should be.

controllers/placementapi_controller.go
gibizer: Thanks!

gibizer:
/test placement-operator-build-deploy-kuttl
```
Warning: Push failed, retrying in 5s ...
Registry server Address: 
Registry server User Name: openstack-k8s-operators+cirobot
Registry server Email: 
Registry server Password: <<non-empty>>
error: build error: Failed to push image: trying to reuse blob sha256:07a64a71e01156f8f99039bc246149925c6d1480d3957de78510bbec6ec68f7a at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": net/http: TLS handshake timeout
```

gibizer (commented):
I'm happy with this. So

api/go.mod
Deydra71: These are unwanted remnants, let me remove them. I also forgot to remove the dependency to https://github.com/openstack-k8s-operators/openstack-operator/pull/625, so I edit it as well.

controllers/placementapi_controller.go
bogdando: Instead of calling ``mgr.GetFieldIndexer()`` multiple times for each [*Field getter](https://github.com/openstack-k8s-operators/placement-operator/pull/92/files#diff-c0f1c404ea10bc348fba06aab6f455b95b7513a337e09b325ebea615fc0ee587R329), you could call it to get the full spec then parse that local copy by the looked up getters.

If mgr.GetFieldIndexer() doesn't allow getting the full spec, you could get ``.spec.tls`` at least

controllers/placementapi_controller.go
bogdando: meant doing the same as you did [above](https://github.com/openstack-k8s-operators/placement-operator/pull/92/files#diff-c0f1c404ea10bc348fba06aab6f455b95b7513a337e09b325ebea615fc0ee587R984)

bogdando (changes requested):

controllers/placementapi_controller.go
Deydra71: I think I understand now your intention. The `IndexField()` function is not for retrieval, but for watching the changes, so storing the spec in a local var would compromise the reconciliation upon changes.

controllers/placementapi_controller.go
Deydra71: I see, but the `deplDef` has not been declared before, apart from `ctrlResult` as used [above](https://github.com/openstack-k8s-operators/placement-operator/pull/92/files#diff-c0f1c404ea10bc348fba06aab6f455b95b7513a337e09b325ebea615fc0ee587R984).

controllers/placementapi_controller.go
bogdando: makes sense for the whole spec, thanks for explaining! what about storing just .spec.tls? That still saves a bunch of redundant requests to kube API server

Deydra71: That might work. I will need take a look if cached data is updated when the `.Spec.TLS` changes. Which will most probably require reconciler to compare the actual  `.Spec.TLS` against the stored one. 

CC @stuggi  what he might think about this.

stuggi: I don't think this will work, but lets investigate any optimization to index information in follow ups as discussed in https://github.com/openstack-k8s-operators/placement-operator/pull/92/files#r1457573645

gibizer (approved):
Looks good to me. I assume it was tested together with https://github.com/openstack-k8s-operators/openstack-operator/pull/625 and works.

controllers/placementapi_controller.go
bogdando: putting this solved to move discussions to the linked thread

bogdando: please make sure to create a follow up

stuggi: I have created https://issues.redhat.com/browse/OSPRH-4208 to keep tracking this

PR #117, opened 2023-12-12
Expose placement-api logs as the pod logs

### LINKED ISSUE #65: Setup logging with side car container
To avoid mixing httpd and placement-api service logs we should configure placement-api to log to a file and add a sidecar container to the deployment to tail that file into its stdout. (See how nova-api does that)

mrkisaolamb:


templates/placementapi/config/logging.conf
```diff
@@ -26,6 +26,12 @@ level=WARNING
 formatter=normal
 args=(sys.stdout,)
 
+[handler_fileHandler]
```
bogdando: This looks different to that we do in other places to get the side car logging.
If that intentional?

templates/placementapi/config/logging.conf
```diff
@@ -26,6 +26,12 @@ level=WARNING
...
+class=FileHandler
+level=WARNING
```
bogdando: what would be a way to set DEBUG?

SeanMooney: this is i guess another pices of tech debt
we fully removed the longing conf form nova we should do that here also.

the correct way to enable debug logging is via the placement.conf

the problem with the placement operator is we still have not rewitten its config generation.
that was a task we had lining up for some for this cycle but then eded up not working on the placement operator at all.

templates/placementapi/config/logging.conf
SeanMooney, line 1: lets chat about this later but if you are reworking logging we should be removing this file and using oslo.log to log to a file with file rotation as we have in nova-api and metadata pods.

templates/placementapi/config/logging.conf
SeanMooney: lets just set this to debug level for now in the logger config and we can then control debug level via the customerServiceConfig

we should be defaulting to debug.

templates/placementapi/config/logging.conf
SeanMooney: for the sake of includign this in Q4 and the next dev preview lets defer the refactor and proceed with the minimal fix to expose the logs for now

SeanMooney (changes requested):
can you update this to enabel debug logging and include them in the file exposed by the side car

templates/placementapi/config/logging.conf
mrkisaolamb: done and also set debug = true in placement.conf

templates/placementapi/config/logging.conf
SeanMooney: we will resolve that in q1 next year as part of proting the config gen as we will remove this file in that work

templates/placementapi/config/placement.conf
```diff
@@ -1,6 +1,7 @@
 [DEFAULT]
 log_config_append = /etc/placement/logging.conf
```
SeanMooney: did you forget to add debug=true here

https://docs.openstack.org/placement/latest/configuration/config.html#DEFAULT.debug

SeanMooney (changes requested):

templates/placementapi/config/placement.conf
mrkisaolamb: even more, i forgot to push chages with --force

gibizer (commented):
There is no placement-api log as the sidecar don't see any logfiles
```
❯ oc logs -c placement-log placement-5b886d4ccf-s9qpl
tail: cannot open '/var/log/placement/placement-api.log' for reading: No such file or directory
```

```
❯ oc logs -c placement-api placement-5b886d4ccf-s9qpl
+ sudo -E kolla_set_configs
sudo: unable to send audit message: Operation not permitted
INFO:__main__:Loading config file at /var/lib/kolla/config_files/config.json
INFO:__main__:Validating config file
INFO:__main__:Kolla config strategy set to: COPY_ALWAYS
INFO:__main__:Copying service configuration files
INFO:__main__:Deleting /etc/placement/placement.conf
INFO:__main__:Copying /var/lib/config-data/merged/placement.conf to /etc/placement/placement.conf
...
```

gibizer (approved):
Looks good. When kuttl is green we can land this

tests/kuttl/common/assert_sample_deployment.yaml
```diff
@@ -101,6 +101,60 @@ spec:
...
+        - -F
+        - /var/log/placement/placement-api.log
+        command:
+        - /usr/bin/dumb-init
+        env:
+        - name: CONFIG_HASH
+          value: n67chfch95hc9h67ch5dbh5cch97h5c9h5d8h66bh89h54hd6h669h555hcch566hb4h4h64bh4hf7hc9hfdh5cfhb8h5fh55h5f5h5cfhbbq
+        - name: KOLLA_CONFIG_STRATEGY
+          value: COPY_ALWAYS
+        image: quay.io/podified-antelope-centos9/openstack-placement-api@sha256:9bcfea1d6ff9ad44243a744cf9e4228e19a83a8eade1161de7fb5fd848befd5d
```
gibizer: I would rather ditch the mentioning of the image hash as it will change independently of the operator code

gibizer (approved):

PR #119, opened 2023-12-15
Set DeploymentReady to False when ReadyCount will be 0

### LINKED ISSUE #69: Make sure that DeploymentReady condition is set back to False if the Deployment ReadyReplicas falls back to 0
https://github.com/openstack-k8s-operators/placement-operator/blob/0f9d8374b3977c1228a41bbce5651f14bd3d2ce2/controllers/placementapi_controller.go#L709-L711

mrkisaolamb:
Make sure that DeploymentReady condition is set back to False if the Deployment ReadyReplicas falls back to 0

### DIFF
--- a/controllers/placementapi_controller.go
+++ b/controllers/placementapi_controller.go
@@ -764,9 +764,19 @@ func (r *PlacementAPIReconciler) reconcileNormal(ctx context.Context, instance *
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 {
+	if instance.Status.ReadyCount > 0 || *instance.Spec.Replicas == 0 {
 		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
+	} else {
+		Log.Info("Deployment is not ready")
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.DeploymentReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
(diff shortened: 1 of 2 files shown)


bogdando (commented):
lgtm

controllers/placementapi_controller.go
```diff
@@ -766,7 +766,17 @@ func (r *PlacementAPIReconciler) reconcileNormal(ctx context.Context, instance *
...
 		instance.Status.Conditions.MarkTrue(condition.DeploymentReadyCondition, condition.DeploymentReadyMessage)
+	} else {
+		Log.Info("Deployment is not ready")
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.DeploymentReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.DeploymentReadyRunningMessage))
+		// It is OK to return success as we are watching for StatefulSet changes
+		return ctrl.Result{}, nil
```
abays, line 777: What if `replicas` is set to `0`?  That might prevent the CR from ever reaching the "ready" state [1].

[1] https://github.com/openstack-k8s-operators/placement-operator/blob/4daf37ecb2bb584e1f3a98b7f1660b81519f8ff1/api/v1beta1/placementapi_types.go#L196-L199

gibizer: @abays you are right. If the replicas is 0 then we can be ready with 0 running pods.

mrkisaolamb: done

PR #129, opened 2024-01-10
Watch input secret and update deployment if a relevant password changes

### LINKED ISSUE #66: Only set InputReady True if the expected fields exists in the Secret
Today InputReady is set if the Secret exists but the expected fields in the Secret is not checked.

### LINKED ISSUE #75: Watch input secret and update deployment if a relevant password changes
As shown in https://github.com/openstack-k8s-operators/placement-operator/pull/74 this does not work today.

mrkisaolamb:


controllers/placementapi_controller.go
```diff
@@ -207,6 +333,8 @@ func (r *PlacementAPIReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.ServiceAccount{}).
 		Owns(&rbacv1.Role{}).
 		Owns(&rbacv1.RoleBinding{}).
+		Watches(&source.Kind{Type: &corev1.Secret{}},
+			handler.EnqueueRequestsFromMapFunc(r.GetSecretMapperFor(&placementv1.PlacementAPIList{}, context.TODO()))).
```
gibizer, line 337: This is in conflict with https://github.com/openstack-k8s-operators/placement-operator/pull/92/files#diff-c0f1c404ea10bc348fba06aab6f455b95b7513a337e09b325ebea615fc0ee587R222 

As far as I understand the pattern for watching for Secret objects in https://github.com/openstack-k8s-operators/placement-operator/pull/92 is more efficient. 

So we need to figure out a way forward that incorporate both changes as this PR has ensureSecret which is good and #92 has an advanced Secret watching solution.

tests/functional/placementapi_controller_test.go
```diff
@@ -136,7 +136,7 @@ var _ = Describe("PlacementAPI controller", func() {
 				names.PlacementAPIName,
 				ConditionGetterFunc(PlacementConditionGetter),
 				condition.InputReadyCondition,
-				corev1.ConditionTrue,
+				corev1.ConditionFalse,
```
gibizer, line 173: nice!

controllers/placementapi_controller.go
```diff
@@ -58,11 +62,133 @@ import (
...
 
+type conditionUpdater interface {
+	Set(c *condition.Condition)
+	MarkTrue(t condition.Type, messageFormat string, messageArgs ...interface{})
+}
+
+type GetSecret interface {
+	GetSecret() string
+	client.Object
+}
```
gibizer, line 73: If we want to move ensureSecret to lib-common then I agree to have this generalization here and the as a second step moved to lib-common. However if we won't do that move then this generalization is unnecessary in placement as in placement we have a single CR type to handle so we can inline that type directly. (In nova we use this generalization as we have multiple CR types to handle with the same code)

controllers/placementapi_controller.go
gibizer: // later
It seem the two watch can coexists based on the proposal in https://github.com/openstack-k8s-operators/nova-operator/pull/646/files#diff-acc4b831bcb31906497644cbef12aa82d90a4a6690078596ca10437735c05e11R850-R856

controllers/placementapi_controller.go
```diff
@@ -591,7 +719,14 @@ func (r *PlacementAPIReconciler) reconcileNormal(ctx context.Context, instance *
 	//
 	// check for required OpenStack secret holding passwords for service/admin user and add hash to the vars map
 	//
-	ospSecret, hash, err := oko_secret.GetSecret(ctx, helper, instance.Spec.Secret, instance.Namespace)
+	hash, result, ospSecret, err := ensureSecret(
+		ctx,
+		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
+		[]string{
+			instance.Spec.PasswordSelectors.Service,
```
gibizer, line 726: We use other fields from the secret too:
https://github.com/openstack-k8s-operators/placement-operator/blob/269c641305039b0f80edb948b6910c92fb3ce7d1/pkg/placement/deployment.go#L167-L168

mrkisaolamb: Thanks, done!

controllers/placementapi_controller.go
mrkisaolamb: Agree, I created issues for both repos https://github.com/openstack-k8s-operators/nova-operator/issues/649 and https://github.com/openstack-k8s-operators/placement-operator/issues/133 . Both I will fix in followup

controllers/placementapi_controller.go
mrkisaolamb: Agree that we can try to incorporate both do we want to create issu for that? IMO it will be good to fix before GA but after TLS support will land in repos

controllers/placementapi_controller.go
gibizer: ack

controllers/placementapi_controller.go
gibizer: Let land this independently from TLS. And after both landed we can try to combine the two watch

gibizer (approved):
Looks good. Thanks!

PR #130, opened 2024-01-11
Use instance.Name in creating resources

### LINKED ISSUE #67: Make sure that Resources created by the placement-operator are named based on the name of the PlacementAPI CR
This is needed to allow two PlacementAPI CR to potentially coexists in the same namespace.
I see two places where this is not followed in the placement-operator:
* The DB Sync Job name https://github.com/openstack-k8s-operators/placement-operator/blob/0f9d8374b3977c1228a41bbce5651f14bd3d2ce2/pkg/placement/dbsync.go#L55
* The Deployment name https://github.com/openstack-k8s-operators/placement-operator/blob/0f9d8374b3977c1228a41bbce5651f14bd3d2ce2/pkg/placement/deployment.go#L91

mrkisaolamb:
Resources created by the placement-operator are
named based on the name of the PlacementAPI CR

### DIFF
--- a/pkg/placement/dbsync.go
+++ b/pkg/placement/dbsync.go
@@ -46,7 +46,7 @@ func DbSyncJob(
 
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      ServiceName + "-db-sync",
+			Name:      instance.Name + "-db-sync",
 			Namespace: instance.Namespace,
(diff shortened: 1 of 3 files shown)


SeanMooney (approved):
yep this is in line with pattern we have established in the nova operators and elsewhere

the name of the created resources should be derived form the cr name not hardcoded based on the service.

PR #135, opened 2024-01-18
Remove init flow

### LINKED ISSUE #64: Align service config generation

mrkisaolamb:


SeanMooney (dismissed):
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openstack-k8s-operators_placement-operator/135/pull-ci-openstack-k8s-operators-placement-operator-main-placement-operator-build-deploy-kuttl/1752358706336501760#1:build-log.txt%3A10637

The pod security context is now different so we need to remove that or fix it in the kuttl test

controllers/placementapi_controller.go
```diff
@@ -267,61 +270,532 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		return result, err
+	} else if (result != ctrl.Result{}) {
+		return result, nil
+	}
+
+	if (err != nil || result != ctrl.Result{}) {
+		// We can ignore RequeueAfter as we are watching the Service resource
+		// but we have to return while waiting for the service to be exposed
+		return ctrl.Result{}, err
+	}
```
gibizer: The comment here is not reflecting the reality. I think you are waiting for the db sync to finish before doing the deployment.

gibizer: But also I think the whole if is pointless as if there is an error then we returning at L450, if there is a requeue request from the db sync then we are returning at L452. So we will arrive here with an error or a requeue.

templates/placementapi/config/placement-api-config.json
```diff
@@ -2,25 +2,25 @@
     "command": "/usr/sbin/httpd -DFOREGROUND",
     "config_files": [
         {
-            "source": "/var/lib/config-data/merged/placement.conf",
+            "source": "/var/lib/config-data/placement.conf",
             "dest": "/etc/placement/placement.conf",
```
gibizer, line 6: Eventually we want to use /etc/placement/placement.conf.d/ but that can be done separately. Also at that point we want to put an empty file to /etc/placement/placement.conf as the container image put the generated sample config file there.

SeanMooney:
/test placement-operator-build-deploy-kuttl

i think the test is racy and its not waiting long enough for it to scale

we may need to merge https://github.com/openstack-k8s-operators/placement-operator/pull/125 first or  modify how kuttl is run

placement_kuttl_run https://github.com/openstack-k8s-operators/install_yamls/blob/60c1fb7828a91545e52b5c799162bcc21c3dd5ee/Makefile#L1586-L1588

its currently looking for kuttl config in repo in https://github.com/openstack-k8s-operators/placement-operator/blob/main/kuttl-test.yaml

so we can adjust some of the config that way

nova is using a timeout of 300 seconds per test

https://github.com/openstack-k8s-operators/nova-operator/blob/main/test/kuttl/test-suites/default/config.yaml#L7

we are using 180 in placement so we could start by bumping to 300 and see if that is enough to stableise the test

controllers/placementapi_controller.go
mrkisaolamb: Done, Thanks!

templates/placementapi/config/placement-api-config.json
mrkisaolamb: ack, should I create new issue or this will be handle as a part of different feature/issue?

mrkisaolamb:
> /test placement-operator-build-deploy-kuttl
> 
> i think the test is racy and its not waiting long enough for it to scale
> 
> we may need to merge #125 first or modify how kuttl is run
> 
> placement_kuttl_run https://github.com/openstack-k8s-operators/install_yamls/blob/60c1fb7828a91545e52b5c799162bcc21c3dd5ee/Makefile#L1586-L1588
> 
> its currently looking for kuttl config in repo in https://github.com/openstack-k8s-operators/placement-operator/blob/main/kuttl-test.yaml
> 
> so we can adjust some of the config that way
> 
> nova is using a timeout of 300 seconds per test
> 
> https://github.com/openstack-k8s-operators/nova-operator/blob/main/test/kuttl/test-suites/default/config.yaml#L7
> 
> we are using 180 in placement so we could start by bumping to 300 and see if that is enough to stableise the test

Thanks! I bumped it to 300 because this failed already few times on same issue that db sync didn't finished

mrkisaolamb:
kuttl tests catch realll issue with volumes for db-sync so this is not related to timeout

pkg/placement/volumes.go
```diff
@@ -63,55 +53,29 @@ func getVolumes(name string) []corev1.Volume {
...
-	return []corev1.VolumeMount{
-		{
-			Name:      "scripts",
-			MountPath: "/usr/local/bin/container-scripts",
-			ReadOnly:  true,
-		},
 		{
-			Name:      "config-data-merged",
+			Name:      "config-data",
 			MountPath: "/var/lib/config-data/merged",
```
gibizer: lets do a separate follow up where the merged postfix is removed from the mount path. That naming is the from the past where the init container merged configuration.

tests/kuttl/common/assert_sample_deployment.yaml
```diff
@@ -212,17 +186,6 @@ status:
...
   replicas: 1
 ---
-# the openshift annotations can't be checked through the deployment above
-apiVersion: v1
-kind: Pod
-metadata:
-  annotations:
-    openshift.io/scc: anyuid
-  labels:
-    service: placement
```
gibizer: Why we dropped the Pod assert? I guess we still have the same placement api pod running

templates/placementapi/config/placement-api-config.json
gibizer: please either create an issue or a quick follow up PR

tests/kuttl/common/assert_sample_deployment.yaml
mrkisaolamb: done

pkg/placement/volumes.go
mrkisaolamb: done

templates/placementapi/config/placement-api-config.json
mrkisaolamb: done in 826dafd93ed358771fe1cf8fbab2006c75e2c89e

gibizer (approved):

PR #140, opened 2024-01-26
[go.mod]Bump from v0.3.0 to pseudoversion

gibizer:
Our intention is to track service operator and lib-common dependencies via pseudoversions. However renovate does not automatically bump from a tagged version (e.g. v0.3.0) to a newer but not tagged pseudoversion. So this patch does the manual bump. After this renovate will bump the newer pseduoversions.


mrkisaolamb:
/retest
fail looks like something not related to this bump

gibizer:
I think we need to manual adapt at least due to the need of the new RBAC rule for MariaDBAccount CR (see https://github.com/openstack-k8s-operators/nova-operator/pull/660)

gibizer:
We can do this bump and do a separate bump with adaptation to the latest mariadb

mrkisaolamb (approved):
lgtm

PR #146, opened 2024-02-07
Bump to latest mariadb-operator/api

gibizer:
Adapt to the new version by:

* adding RBAC rules for MariaDBAccount manipulation
* adding Own(MariaDBAccount) to detect status changes
* separating PlacementAPI CR name from the schema and
  MariaDBDatabase/Account name
* adapting test to simulate MariaDBAccount success
* adapting to new DB hostnames due to the mariadb TLS feature

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -137,6 +137,24 @@ rules:
 - apiGroups:
   - mariadb.openstack.org
   resources:
+  - mariadbaccounts
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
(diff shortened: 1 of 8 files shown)


mrkisaolamb (approved):
lgtm

SeanMooney (approved):
my understaidn of this change is as follows.

we are bumping the operator version which provides the new crd definition fo the mariadb accounts. we have adapted to the new API function and added an owner ref to the mariadb resource. we created.

the kuttle tests have been updated to reflect that we now use the cluster local service name and is now explicitly using the fqdn instead of relying on the implicit search domain.

with that context and the assert that accepting a mariadb account is still a follow up for a different PR in our crd this looks correct to me

PR #150, opened 2024-02-14
Revert "Controller runtime update"

gibizer:
Reverts openstack-k8s-operators/placement-operator#144

There is no ubi9/go-toolset:1.21 available and won't be until May. So we can only move to controller-runtime 0.16 and golang 1.20.


gibizer:
We are going back to a known good state with this revert so I'm merging it without waiting for CI

PR #165, opened 2024-02-19
[tlse] TLS database connection

stuggi:
moves requesting the DB before rendering the service configuration. The my.cnf file gets added to the secret holding the service configs. The content of my.cnf is centrally managed in the mariadb-operator and retrieved calling db.GetDatabaseClientConfig(tlsCfg)

Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/190
Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/191

Jira: [OSPRH-4547](https://issues.redhat.com//browse/OSPRH-4547)

### DIFF
--- a/controllers/placementapi_controller.go
+++ b/controllers/placementapi_controller.go
@@ -349,7 +349,14 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	// all our input checks out so report InputReady
 	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.InputReadyMessage)
 
-	err = r.generateServiceConfigMaps(ctx, h, instance, secret, &configMapVars)
+	db, result, err := r.ensureDB(ctx, h, instance)
+	if err != nil {
+		return ctrl.Result{}, err
+	} else if (result != ctrl.Result{}) {
+		return result, nil
+	}
+
+	err = r.generateServiceConfigMaps(ctx, h, instance, secret, &configMapVars, db)
 	if err != nil {
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			condition.ServiceConfigReadyCondition,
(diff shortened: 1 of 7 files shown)


controllers/placementapi_controller.go
```diff
@@ -349,6 +349,13 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	// all our input checks out so report InputReady
 	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.InputReadyMessage)
 
+	result, err = r.ensureDB(ctx, h, instance)
+	if err != nil {
+		return ctrl.Result{}, err
+	} else if (result != ctrl.Result{}) {
+		return result, nil
+	}
+
```
gibizer, line 358: thanks this fixing a bug

controllers/placementapi_controller.go
```diff
@@ -1184,10 +1186,19 @@ func (r *PlacementAPIReconciler) generateServiceConfigMaps(
 
 	cmLabels := labels.GetLabels(instance, labels.GetGroupLabel(placement.ServiceName), map[string]string{})
 
+	db, err := mariadbv1.GetDatabaseByName(ctx, h, placement.DatabaseName)
```
gibizer: I think we could return the db from ensureDB instead

templates/placementapi/config/placement-api-config.json
```diff
@@ -47,6 +47,12 @@
             "owner": "placement",
             "perm": "0600",
             "optional": true
+        },
+        {
+            "source": "/var/lib/openstack/config/my.cnf",
+            "dest": "/etc/my.cnf",
+            "owner": "root",
```
gibizer: why not placement? as for the placenent.conf

stuggi: I can change that,but its a file which could be read by anyone, so its not really an issue who owns it. I personally tend to have files owned by root if they can be read by anyone. I'll change it

controllers/placementapi_controller.go
stuggi: yes, will do that

templates/placementapi/config/placement-api-config.json
gibizer: does not matter much but we can also make it 0600 to only be readable by placemnet

gibizer (approved):
looks good

PR #167, opened 2024-02-23
migrate from databaseUsername to databaseAccount

zzzeek:
This moves placement to fully use MariaDBAccount based on the dev work being done for mariadb-operator:

https://github.com/openstack-k8s-operators/mariadb-operator/pull/184/files

Lead Jira: [OSPRH-4095](https://issues.redhat.com/browse/OSPRH-4095)

1. [x] controller calls EnsureMariaDBAccount up front to make sure MariaDBAccount is present
2. [x] error code from EnsureMariaDBAccount is handled, Conditions are amended when error is returned
3. [x] controller calls NewDatabaseForAccount instead of NewDatabase
4. [x]  GetAccountAndSecret is used to retrieve account /secret to populate template
5. [x]  GetDatabaseByName() , normally used for delete finalizers, replaced with GetDatabaseByNameAndAccount
6. [x]  CreateOrPatchAll() used to patch the Database, replacing CreateOrPatchDB / CreateOrPatchDBByName
7. [x]  controller calls DeleteUnusedMariaDBAccountFinalizers when launched pods are definitely running on a new MariaDBAccount, returns error code if present
8. [x]  PasswordSelectors that refer to database are removed
9. [x]  all databaseUser replaced with databaseAccount inside of all XYZ_types.go
10. [x] all databaseUser replaced with databaseAccount inside of all `kuttl/*.yaml`
11. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all XYZ_types.go
12. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all `kuttl/*.yaml`
13. [x] MariaDBAccountSuiteTests are used in controller ginkgo tests if it has them
14. [x] Use configsecrets for database URLs; remove from job hash  - already was like that
15. [x] [184](https://github.com/openstack-k8s-operators/mariadb-operator/pull/184) is merged and replaces from go.mod are removed

### DIFF
--- a/controllers/placementapi_controller.go
+++ b/controllers/placementapi_controller.go
@@ -323,7 +323,6 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
 		[]string{
 			instance.Spec.PasswordSelectors.Service,
-			instance.Spec.PasswordSelectors.Database,
 		},
 		h.GetClient(),
 		&instance.Status.Conditions)
@@ -348,6 +347,32 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	// all our input checks out so report InputReady
 	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.InputReadyMessage)
+
+	// ensure MariaDBAccount exists.  This account record may be created by
+	// openstack-operator or the cloud operator up front without a specific
+	// MariaDBDatabase configured yet.   Otherwise, a MariaDBAccount CR is
+	// created here with a generated username as well as a secret with
+	// generated password.   The MariaDBAccount is created without being
+	// yet associated with any MariaDBDatabase.
+	_, _, err = mariadbv1.EnsureMariaDBAccount(
+		ctx, h, instance.Spec.DatabaseAccount,
+		instance.Namespace, false, placement.DatabaseName,
+	)
+
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			mariadbv1.MariaDBAccountReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			mariadbv1.MariaDBAccountNotReadyMessage,
+			err.Error()))
+
+		return ctrl.Result{}, err
+	}
+	instance.Status.Conditions.MarkTrue(
+		mariadbv1.MariaDBAccountReadyCondition,
+		mariadbv1.MariaDBAccountReadyMessage,
+	)
 
 	db, result, err := r.ensureDB(ctx, h, instance)
 	if err != nil {
@@ -450,6 +475,7 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		// We can ignore RequeueAfter as we are watching the KeystoneEndpoint resource
 		return ctrl.Result{}, err
 	}
+
 	result, err = r.ensureDbSync(ctx, instance, h, serviceAnnotations)
 	if (err != nil || result != ctrl.Result{}) {
 		return result, err
@@ -464,6 +490,12 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	if !instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition) {
 		Log.Info("Waiting for the Deployment to become Ready before exposing the sevice in Keystone")
 		return ctrl.Result{}, nil
+	}
+
+	// remove finalizers from unused MariaDBAccount records
+	err = mariadbv1.DeleteUnusedMariaDBAccountFinalizers(ctx, h, placement.DatabaseName, instance.Spec.DatabaseAccount, instance.Namespace)
+	if err != nil {
+		return ctrl.Result{}, err
 	}
 
 	return ctrl.Result{}, nil
@@ -919,7 +951,7 @@ func (r *PlacementAPIReconciler) reconcileDelete(ctx context.Context, instance *
 	Log.Info("Reconciling Service delete")
 
 	// remove db finalizer before the placement one
-	db, err := mariadbv1.GetDatabaseByName(ctx, helper, placement.DatabaseName)
+	db, err := mariadbv1.GetDatabaseByNameAndAccount(ctx, helper, placement.DatabaseName, instance.Spec.DatabaseAccount, instance.Namespace)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return ctrl.Result{}, err
 	}
@@ -974,23 +1006,16 @@ func (r *PlacementAPIReconciler) ensureDB(
 	h *helper.Helper,
 	instance *placementv1.PlacementAPI,
 ) (*mariadbv1.Database, ctrl.Result, error) {
-	// (ksambor) should we use  NewDatabaseWithNamespace instead?
-	db := mariadbv1.NewDatabaseWithNamespace(
-		placement.DatabaseName,
-		instance.Spec.DatabaseUser,
-		instance.Spec.Secret,
-		map[string]string{
-			"dbName": instance.Spec.DatabaseInstance,
-		},
-		placement.DatabaseName,
-		instance.Namespace,
+	db := mariadbv1.NewDatabaseForAccount(
+		instance.Spec.DatabaseInstance, // mariadb/galera service to target
+		placement.DatabaseName,         // name used in CREATE DATABASE in mariadb
+		placement.DatabaseName,         // CR name for MariaDBDatabase
+		instance.Spec.DatabaseAccount,  // CR name for MariaDBAccount
+		instance.Namespace,             // namespace
 	)
+
 	// create or patch the DB
-	ctrlResult, err := db.CreateOrPatchDBByName(
-		ctx,
-		h,
-		instance.Spec.DatabaseInstance,
-	)
+	ctrlResult, err := db.CreateOrPatchAll(ctx, h)
 	if err != nil {
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			condition.DBReadyCondition,
@@ -1216,6 +1241,10 @@ func (r *PlacementAPIReconciler) generateServiceConfigMaps(
 	if err != nil {
 		return err
 	}
+
+	databaseAccount := db.GetAccount()
+	dbSecret := db.GetSecret()
+
 	templateParameters := map[string]interface{}{
 		"ServiceUser":         instance.Spec.ServiceUser,
 		"KeystoneInternalURL": keystoneInternalURL,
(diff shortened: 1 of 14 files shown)


controllers/placementapi_controller.go
```diff
@@ -441,7 +466,7 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	err = r.ensureKeystoneServiceUser(ctx, h, instance)
-	if err != nil {
+	if (err != nil || result != ctrl.Result{}) {
```
zzzeek: oh this is obsolete now, merge artifact

tests/functional/placementapi_controller_test.go
```diff
@@ -371,6 +380,33 @@ var _ = Describe("PlacementAPI controller", func() {
 			Expect(binding.Subjects).To(HaveLen(1))
 			Expect(binding.Subjects[0].Name).To(Equal(sa.Name))
 		})
+
+		It("creates MariaDB database", func() {
```
zzzeek: whats this, merge artifact?  fix that....

tests/functional/base_test.go
```diff
@@ -60,6 +61,9 @@ func CreateNames(placementAPIName types.NamespacedName) Names {
 		MariaDBDatabaseName: types.NamespacedName{
 			Namespace: placementAPIName.Namespace,
 			Name:      placement.DatabaseName},
+		MariaDBAccount: types.NamespacedName{
+			Namespace: placementAPIName.Namespace,
+			Name:      "placement"},
```
zzzeek: dont hardcode this, there are mistakes in the test suite because this is the same as database name.  use a variable with a distinct name

controllers/placementapi_controller.go
```diff
@@ -349,6 +348,32 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
+	// ensure MariaDBAccount exists.  This account record may be created by
+	// openstack-operator or the cloud operator up front without a specific
+	// MariaDBDatabase configured yet.   Otherwise, a MariaDBAccount CR is
+	// created here with a generated username as well as a secret with
+	// generated password.   The MariaDBAccount is created without being
+	// yet associated with any MariaDBDatabase.
+	_, _, err = mariadbv1.EnsureMariaDBAccount(
+		ctx, h, instance.Spec.DatabaseAccount,
+		instance.Namespace, false, placement.DatabaseName,
```
gibizer, line 359: do we need to set the requireTLS flag based on some input?

controllers/placementapi_controller.go
```diff
@@ -349,6 +348,32 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	// all our input checks out so report InputReady
 	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.InputReadyMessage)
 
+	// ensure MariaDBAccount exists.  This account record may be created by
+	// openstack-operator or the cloud operator up front without a specific
+	// MariaDBDatabase configured yet.   Otherwise, a MariaDBAccount CR is
+	// created here with a generated username as well as a secret with
+	// generated password.   The MariaDBAccount is created without being
+	// yet associated with any MariaDBDatabase.
+	_, _, err = mariadbv1.EnsureMariaDBAccount(
```
gibizer, line 357: I feel that ensuring the DBAccount logically belongs to ensureDB, so I would move this new logic into that func.

tests/functional/placementapi_controller_test.go
```diff
@@ -843,6 +853,70 @@ var _ = Describe("PlacementAPI controller", func() {
...
 		})
 	})
+
+	// Run MariaDBAccount suite tests.  these are pre-packaged ginkgo tests
+	// that exercise standard account create / update patterns that should be
+	// common to all controllers that ensure MariaDBAccount CRs.
+	mariadb_test.MariaDBAccountSuiteTests(
+		"Placememt",
+		// Populate test variables which will run inside of BeforeEach
+		func(harness *mariadb_test.MariaDBTestHarness) {
```
gibizer: As golang does not have keyword args these functions does not have a name here to signify their purpose and 
 their requirements. Would it be more readable to pass a struct with func fields to MariaDBAccountSuiteTests instead? Or does it violate some assumption within the ginkgo execution model itself?

tests/functional/placementapi_controller_test.go
```diff
@@ -843,6 +853,70 @@ var _ = Describe("PlacementAPI controller", func() {
...
+				g.Expect(th.K8sClient.Update(ctx, placementapi)).Should(Succeed())
+			}, timeout, interval).Should(Succeed())
+
+		},
+		// delete the keystone instance to exercise finalizer removal
+		func() {
+			th.DeleteInstance(GetPlacementAPI(names.PlacementAPIName))
+		},
+	)
+
```
gibizer, line 936: Could you add a test that shows that when the account is changed the service config is regenerated with the new user and password and the hash in the statefulset's env is updated to trigger pod restarts?

controllers/placementapi_controller.go
stuggi: tls is used as soon the mariadbdatabase reflects it is hosted on a galera supporting tls and the service has a CA to validate it. we could add setting requiredTLS as a follow up. maybe after we switched our default to enable podlevel/internal tls per default?

gibizer: I'm OK about a follow up, just seemed strange that we merged #165  yesterday in placement, and here we explicitly disable it. So I wanted to make sure it is intentional.

controllers/placementapi_controller.go
zzzeek: Yes I noticed that the recent merge has consolidated DB functions to be ahead of configmaps in any case, which was one of the reasons ensureaccount was up there.   since ensureaccount isn't actually going away, for those controllers where ensuredb is before configmaps (Which I guess will be all of them now), might as well put ensure account into it.    it starts to look like the whole of the various ensureDB functions could just be one big function in mariadb-operator, but i dont know that we have time to do all that just yet.

tests/functional/placementapi_controller_test.go
zzzeek: not sure?  I was just doing this whatever way I could get the funcs passed along, what benefits does a struct offer?   the thing about populating structs is that it seems you can leave fields omitted which is what I dont want here, every field is required.

tests/functional/placementapi_controller_test.go
zzzeek: those assertions would be specific to placement here.   this would either be a test that's specific to the placement repo (and also the other repos that it becomes a part of), or maybe the mariadb harness here could offer a hook for additional suites based on an account change that leverages the existing fixture functions.

controllers/placementapi_controller.go
zzzeek: the ensuremariadbaccount is ultimately going to be a receiver of a MariaDBAccount that may be coming from openstack-operator, and the account is not associated with any database as of yet.   at this stage, it's not appropriate to set TLS, because nothing is yet known about what kind of database this account will be applied to.     if we are flipping a TLS flag on when we do the ensure MariaDBDatabase phase, then we would want to flip on "require TLS" on the MariaDBAccount as well, and that would be within the mariadbdatabase_funcs.   if that's the use we want?  that is, if DB supports TLS, then all accounts should *require* TLS.  is that a req?

stuggi: > I'm OK about a follow up, just seemed strange that we merged #165 yesterday in placement, and here we explicitly disable it. So I wanted to make sure it is intentional.

right, setting requireTLS == false is not that then no tls is used for db connection. it is just that this db user could use tls and non tls to connect. if tls is used is being controlled by the my.cnf content. 

>  that is, if DB supports TLS, then all accounts should _require_ TLS. is that a req?

I am not sure right now. I can discuss it in our tomorrows tls meeting and get back

tests/functional/placementapi_controller_test.go
zzzeek: OK struct has names.   So.... make one struct and send to the function, how about this?  give this the thumbs up and I can change to this, I added a declare/run/go/???? method which takes the helper and timeout/interval


```
	// Run MariaDBAccount suite tests.  these are pre-packaged ginkgo tests
	// that exercise standard account create / update patterns that should be
	// common to all controllers that ensure MariaDBAccount CRs.
	mariadb_suite := &mariadb_test.MariaDBAccountSuite{
		name:          "Placement",
		namespace:     names.PlacementAPIName.Namespace,
		databaseName:  placement.DatabaseName,
		finalizerName: "PlacementAPI",
		generateCR: func(accountName types.NamespacedName) {
			// ...
		},
		updateAccount: func(newAccountName types.NamespacedName) {

			Eventually(func(g Gomega) {
				placementapi := GetPlacementAPI(names.PlacementAPIName)
				placementapi.Spec.DatabaseAccount = newAccountName.Name
				g.Expect(th.K8sClient.Update(ctx, placementapi)).Should(Succeed())
			}, timeout, interval).Should(Succeed())

		},
		deleteCR: func() {
			th.DeleteInstance(GetPlacementAPI(names.PlacementAPIName))
		},
	}
        // DeclareGinkgo() ?  Run()?  
	mariadb_suite.DeclareSuite(
		mariadb, timeout, interval,
	)

```

controllers/placementapi_controller.go
gibizer: thanks for the explanation. I defer to @stuggi if this is OK as is.

controllers/placementapi_controller.go
gibizer: yes, @stuggi fixed the ordering of the steps to have db created before we generate config. I would do the ensureDB consolidation here locally, and maybe looking at a common mariadb-operator/api function later separatly

tests/functional/placementapi_controller_test.go
gibizer: on the optionality of the struct fields, that is a downside here yes, but you can add a runtime check for it.

gibizer: my soft preference is towards mariadb_suite.Run()

tests/functional/placementapi_controller_test.go
gibizer: I would go with the placement specific test case as that seems simpler for this than extra hooks in the harness.

zzzeek: I added additional helper suites in mariadb so here we can see it's doing just the placement specific parts of it.  this is done as an add-on as in other projects the URL / confighash can be broken up based on controller and whatnot.

gibizer (commented):
looks good to me. We just need to land the mariadb dependency first

zzzeek:
> looks good to me. We just need to land the mariadb dependency first

yah once mariadb lands I can re-crank through everyone and add mistral / telemetry / barbican also

tests/functional/placementapi_controller_test.go
zzzeek: I added additional "optional / run multiple times" suites to it as well so it now has `RunBasicSuite()` `RunURLAssertSuite(assertURL assertsURL)` `RunConfigHashSuite(getConfigHash getsConfigHash)` 

i think this makes it a lot easier to test everyone's ability to render expected URLs upfront/on change as well as being able to assert any number of confighashes that are expected to change

controllers/placementapi_controller.go
stuggi: we did not came up with a real preference if "require TLS" should be used as soon the DB supports TLS. since the services will use tls as soon the DB supports it and the service has the CA we could set it, but atm it won't hurt keep it available. I'd personally do it in a follow up, after we tested all the different scenarios, also with adoption in mind. adoption of tlse env is still in early phase.

zzzeek: yah i think this should be all followup stuff.  we're just trying to get all the levers available right now (like, every controller gets a distinct DB password, stuff like that :)  )

gibizer (approved):
I tested this and I was able to rotate the db password properly. Nice work!

PR #169, opened 2024-03-01
Add PlacementAPISpecCore struct

dprince:
This version of the struct (called "core") is meant to be used via the OpenStackControlplane. It is the same as PlacementAPISpec only it is missing the containerImages.

The Default() function for webhooks has been split accordingly.

Jira: [OSPRH-4835](https://issues.redhat.com//browse/OSPRH-4835)


mrkisaolamb (commented):
I'm ok with this patch, but I will wait for @gibizer review

api/v1beta1/placementapi_webhook.go
```diff
@@ -121,18 +126,27 @@ func (r *PlacementAPI) ValidateDelete() (admission.Warnings, error) {
 }
 
 func (r PlacementAPISpec) ValidateCreate(basePath *field.Path) field.ErrorList {
```
gibizer, line 128: note to myself: This is called by the webhook within the placement-operator via `func (r *PlacementAPI) ValidateCreate()`

api/v1beta1/placementapi_webhook.go
```diff
@@ -121,18 +126,27 @@ func (r *PlacementAPI) ValidateDelete() (admission.Warnings, error) {
...
-	return r.ValidateDefaultConfigOverwrite(basePath)
+	return ValidateDefaultConfigOverwrite(basePath, r.DefaultConfigOverwrite)
 }
 
 func (r PlacementAPISpec) ValidateUpdate(old PlacementAPISpec, basePath *field.Path) field.ErrorList {
-	return r.ValidateDefaultConfigOverwrite(basePath)
+	return ValidateDefaultConfigOverwrite(basePath, r.DefaultConfigOverwrite)
+}
+
+func (r PlacementAPISpecCore) ValidateCreate(basePath *field.Path) field.ErrorList {
```
gibizer, line 136: note to myself: this will be called by the webhook in openstack-operator when validating the service core struct within the openstackcontrolplane struct.

api/v1beta1/placementapi_webhook.go
```diff
@@ -121,18 +126,27 @@ func (r *PlacementAPI) ValidateDelete() (admission.Warnings, error) {
...
+	return ValidateDefaultConfigOverwrite(basePath, r.DefaultConfigOverwrite)
+}
+
+func (r PlacementAPISpecCore) ValidateUpdate(old PlacementAPISpec, basePath *field.Path) field.ErrorList {
+	return ValidateDefaultConfigOverwrite(basePath, r.DefaultConfigOverwrite)
 }
 
-func (r PlacementAPISpec) ValidateDefaultConfigOverwrite(
+func ValidateDefaultConfigOverwrite(
```
gibizer, line 144: An alternative would be to change this to use the PlacementAPISpecCore as a receiver. Then 
* `func (r PlacementAPISpecCore) ValidateCreate` can call it directly as the type matches
* `func (r PlacementAPISpec) ValidateCreate` can call it on its internal core struct within its receiver. Golang actually expose the functions on the embedded struct on the embedding struct automatically.

gibizer: This would reduce the change to https://github.com/gibizer/placement-operator/commit/91e9a074e1300722060710acdda4bc90c349cd82

While this is not a big deal here in placement, I think if we apply this pattern to the nova PR that would reduce the change there significantly

gibizer: Applying it to nova is not a clear cut due to deeper nesting of structs without a common embedded base struct.

gibizer (approved):
Looks OK. I will consider refactoring this to the mentioned pattern later.

PR #171, opened 2024-03-05
use Eventually in RunURLAssertSuite test

zzzeek:
The RunURLAssertSuite harness currently waits on finalizers added to MariaDBAccount as an indication when it's safe to test that the new URL is present in the config.  however, in [1] we are discussing removing the use of finalizers here so that this harness can be generalized to more kinds of controllers. as such, if the assertion here uses Eventually, it becomes compatible with this change, as the waiting for the condition is implemented directly here.

[1]
https://github.com/openstack-k8s-operators/mariadb-operator/pull/205#pullrequestreview-1916137443


gibizer (commented):
Looks good. Regardless of the mariadb change this makes the test more stable

gibizer (approved):

PR #174, opened 2024-03-11
Correct ValidateUpdate for PlacementSpecCore

dprince:
Updates the function so that the type of the "old" and reciever are the same.


gibizer (approved):
Make sense, thanks.

PR #177, opened 2024-03-21
Re-init conditions each reconcile

mrkisaolamb:
Jira: [OSPRH-5698](https://issues.redhat.com//browse/OSPRH-5698)

### DIFF
--- a/controllers/placementapi_controller.go
+++ b/controllers/placementapi_controller.go
@@ -256,6 +256,10 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		Log.Error(err, "Failed to create lib-common Helper")
 		return ctrl.Result{}, err
 	}
+
+	// Save a copy of the condtions so that we can restore the LastTransitionTime
+	// when a condition's state doesn't change.
+	savedConditions := instance.Status.Conditions.DeepCopy()
 	// initialize status fields
 	if err = r.initStatus(ctx, h, instance); err != nil {
 		return ctrl.Result{}, err
@@ -263,6 +267,7 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	// Always patch the instance status when exiting this function so we can persist any changes.
 	defer func() {
+		condition.RestoreLastTransitionTimes(&instance.Status.Conditions, savedConditions)
 		// update the Ready condition based on the sub conditions
 		if instance.Status.Conditions.AllSubConditionIsTrue() {
 			instance.Status.Conditions.MarkTrue(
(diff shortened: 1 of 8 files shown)


bogdando (commented):
lgtm

controllers/placementapi_controller.go
```diff
@@ -750,79 +750,77 @@ func (r *PlacementAPIReconciler) initStatus(
...
-			condition.UnknownCondition(
-				condition.RoleBindingReadyCondition,
-				condition.InitReason,
-				condition.RoleBindingReadyInitMessage),
-			condition.UnknownCondition(
-				condition.TLSInputReadyCondition,
-				condition.InitReason,
-				condition.InputReadyInitMessage),
-		)
+	instance.Status.Conditions = condition.Conditions{}
```
bogdando: don't we want doing this conditionally, like cinder proposed https://github.com/openstack-k8s-operators/cinder-operator/pull/364/files?diff=split&w=1#diff-b57d545aef75db901bea7faa1e1e6fd8e93b88c0dc1638a8bb27258bf530c57aR169

bogdando (dismissed):
some minor concern

controllers/placementapi_controller.go
mrkisaolamb: IMO this will not give us anything (exept that some conditions with status unknown will keep old LastTransitionTime) but we can change this too be consistent

bogdando: I agree it seems there is no need for nil checks

bogdando (commented):
lgtm

controllers/placementapi_controller.go
gibizer: So as far as I see `	instance.Status.Conditions.Init(&cl)` is smart and it updates existing conditions only if there was status change.

However `instance.Status.Conditions = condition.Conditions{}` blows away the previous list of conditions in the Status. So I think I see why cinder did this only once but calls Init unconditionally. 

With this PR a later reconcile:
* deletes the previous condition list
* creates a new condition list with unknown state (no last transition time)
* do the reconciliation and set condition states (conditions with updated state get a transition time)

Cinder's solution:
* keeps the previous condition list
* sets all condition state to Unknown in the list. (If the state wasn't unknown before then this updates the last transition time)
* do the reconciliation and set condition states accordingly (conditions with updated state get a transition time)

So basically both solutions make the last transition time of non Unknown conditions useless as both solution will reset that field at each reconciliation.

Also both solutions are noisy at the end as a later reconciliaton overwrites the transition time of the conditions in the API.

I will bring up loosing the last transition time issue on slack.

gibizer (approved):
API def looks good and the the implementation in sync with the rest of the services. We will improve the impl later separately

PR #190, opened 2024-04-24
Bump Placement probe timeouts

SeanMooney:
When deploying locally in crc placement can crashloop
indefintly because the readiness and liveness probe
were set to strictly.

The nova-operator relaxed the probes becuase of tempest failures
in openstack-k8s-operators/nova-operator#381

This change updates the placement values to match.


mrkisaolamb (approved):
lgtm

PR #195, opened 2024-05-07
Check for valid service override endpoint type on create and update

stuggi:
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/505

### DIFF
--- a/api/v1beta1/placementapi_webhook.go
+++ b/api/v1beta1/placementapi_webhook.go
@@ -24,6 +24,7 @@ package v1beta1
 import (
 	"fmt"
 
+	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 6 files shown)


gibizer (commented):
Looks good. Extra thanks for the webhook envtest. Once the dependency lands, and we are out of the freeze I'm OK to land this.

api/v1beta1/placementapi_webhook.go
```diff
@@ -126,19 +127,33 @@ func (r *PlacementAPI) ValidateDelete() (admission.Warnings, error) {
...
+
+	return allErrs
 }
 
 func (r PlacementAPISpecCore) ValidateUpdate(old PlacementAPISpecCore, basePath *field.Path) field.ErrorList {
-	return ValidateDefaultConfigOverwrite(basePath, r.DefaultConfigOverwrite)
+	var allErrs field.ErrorList
+
+	// validate the service override key is valid
+	allErrs = append(allErrs, service.ValidateRoutedOverrides(basePath, r.Override.Service)...)
```
gibizer: I'm wondering if ValidateRoutedOverrides could be more generic if the caller pass a path that is already in sync with the field value passed in. E.g. if in a specific service operator the override service field is somewhere else in the instance then the following patter can keep the lib-common code ignorant about that:
```golang
	allErrs = append(allErrs, service.ValidateRoutedOverrides(basePath.Child("override").Child("service"), r.Override.Service)...)
```

stuggi: done

gibizer (approved):

PR #196, opened 2024-05-13
[pre-commit][kuttl]Check for multiple TestAsserts

gibizer:
Add a new pre-commit check to avoid using more than one TestAssert in a
single kuttl assert file as that can lead to false positives as only the
last TestAssert is run by kuttl.

Fortunately there was no such issue in this repo.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,3 +53,11 @@ repos:
     - id: no-commit-to-branch
     - id: trailing-whitespace
       exclude: ^vendor
+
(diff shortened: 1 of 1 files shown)


SeanMooney (approved):
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_placement-operator/196/pull-ci-openstack-k8s-operators-placement-operator-main-precommit-check/1789999698187128832/artifacts/test/build-log.txt


this also looks good

PR #197, opened 2024-05-15
Correct placement flow to expose service only when deployment is ready

### LINKED ISSUE #188: Correct placement flow to expose service only when deployment is ready
On the end of the loop we had
 `	// Only expose the service is the deployment succeeded
	if !instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition) {
		Log.Info("Waiting for the Deployment to become Ready before exposing the sevice in Keystone")
		return ctrl.Result{}, nil
	}
	`
which doesn't make any sense and mess with correct status order of conditions. So ensureServiceExposed should be moved to correct place and check should be readded

mrkisaolamb:
Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/222


controllers/placementapi_controller.go
```diff
@@ -459,6 +448,21 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+	// Only expose the service is the deployment succeeded
+	if !instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition) {
```
gibizer, line 447: We need to reorder these as well.

The mariadb account finalizer can only be removed if we are sure that the deployment is ready as that is the only way we can be make sure that the deployment is now only using the new account.

I suggest to simply move the DeleteUnusedMariaDBAccountFinalizers call as the last step of the Reconciler.

mrkisaolamb: DeleteUnusedMariaDBAccountFinalizers is now the last thing in reconciler

mrkisaolamb:
This patch hit similarr issue like nova and tests timeouts when wait for correct state in db so I bump up timeout for tests

controllers/placementapi_controller.go
```diff
@@ -433,6 +433,21 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+
+	result, err = r.ensureDeployment(ctx, h, instance, inputHash, serviceAnnotations)
+	if (err != nil || result != ctrl.Result{}) {
+		return result, err
+	}
+
+	// Only expose the service is the deployment succeeded
+	if !instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition) {
+		Log.Info("Waiting for the Deployment to become Ready before exposing the sevice in Keystone")
+		return ctrl.Result{}, nil
```
bogdando, line 449: shouldn't we requeue, or will we get another event to retrigger this, once things get ready?

gibizer: we are watching the Deployment so there will be an automatic reconcile when the deployment status changes.

gibizer (approved):
This is stable for me after 40 run. The logic in the controller looks good. We might want to cleanup the envtest if possible later separately.

PR #199, opened 2024-05-20
Set fips-compliant CRD feature flag

bshephar:
This change adds the feature flag to indicate that our operator is fips compliant.


gibizer:
@bshephar do we need this to Beta or only for GA?

gibizer (approved):

PR #206, opened 2024-05-28
[pre-commit]Run make bundle 

gibizer:
We noticed that make bundle can re-generate some of the manifests
causing local diffs in later stages. So to catch such re-generation
needs lets run it in pre-commit.


mrkisaolamb (approved):
lgtm

PR #210, opened 2024-06-10
update csv for fips-compliant

yazug:
recent pull request put it in different place

this patch was generated with make bundle


gibizer (approved):

gibizer:
Fixed this in #206 along with a check that will prevent it to happen again.

PR #220, opened 2024-06-25
Ensure bundles include all the Required Infrastructure Annotations

mrkisaolamb:
Implements: [OSPRH-7936](https://issues.redhat.com/browse/OSPRH-7936)

### DIFF
--- a/config/manifests/bases/placement-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/placement-operator.clusterserviceversion.yaml
@@ -6,8 +6,12 @@ metadata:
     capabilities: Basic Install
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     operatorframework.io/suggested-namespace: openstack
-    operators.openshift.io/infrastructure-features: '["disconnected"]'
     operators.operatorframework.io/operator-type: non-standalone
   name: placement-operator.v0.0.0
   namespace: placeholder


gibizer (commented):
I cannot comment on the TLS flag. I think we need somebody from security to make a decision. I don't know what are the "well known" tunables, and and also the "any workloads it manages" is a very blanket statement.

config/manifests/bases/placement-operator.clusterserviceversion.yaml
```diff
@@ -6,8 +6,12 @@ metadata:
     capabilities: Basic Install
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "true"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     operatorframework.io/suggested-namespace: openstack
-    operators.openshift.io/infrastructure-features: '["disconnected"]'
```
SeanMooney, line 10: i guess this is superceeded by features.operators.openshift.io/disconnected: "true"

config/manifests/bases/placement-operator.clusterserviceversion.yaml
```diff
@@ -6,8 +6,12 @@ metadata:
     capabilities: Basic Install
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "true"
```
SeanMooney: this one we were a littel unsure about.
the only tls endpoint the operator has is the webhook.

if this is referring to the deployment of placement then in theory you can use the service override to tweak the tls profile of the relevant openshfit route that is used to expose it so in that context it would be true.

did we get clarity on if that counted in this context.

config/manifests/bases/placement-operator.clusterserviceversion.yaml
mrkisaolamb: Yes this is also a deprecated type of annotation https://docs.okd.io/latest/operators/operator_sdk/osdk-generating-csvs.html#osdk-csv-annotations-infra_osdk-generating-csvs

config/manifests/bases/placement-operator.clusterserviceversion.yaml
mrkisaolamb: for a now there is no clarity with tls so I put this pr on hold

config/manifests/bases/placement-operator.clusterserviceversion.yaml
```diff
@@ -6,8 +6,12 @@ metadata:
     capabilities: Basic Install
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
```
gibizer, line 13: this is aligned with what we agreed on the today's podified call.

gibizer (approved):

PR #239, opened 2024-08-06
Return nil error in case NAD not found to use RequeueAfter

stuggi:
Currently the reconciler returned both a non-zero result and a non-nil error.
The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff.

In case of NotFound return nil that the ReqeueAfter is used.
    
For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler

### DIFF
--- a/controllers/placementapi_controller.go
+++ b/controllers/placementapi_controller.go
@@ -296,12 +296,13 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		&instance.Status.Conditions)
 	if err != nil {
 		if k8s_errors.IsNotFound(err) {
+			Log.Info(fmt.Sprintf("OpenStack secret %s not found", instance.Spec.Secret))
 			instance.Status.Conditions.Set(condition.FalseCondition(
 				condition.InputReadyCondition,
 				condition.RequestedReason,
 				condition.SeverityInfo,
 				condition.InputReadyWaitingMessage))
-			return ctrl.Result{RequeueAfter: time.Second * 10}, fmt.Errorf("OpenStack secret %s not found", instance.Spec.Secret)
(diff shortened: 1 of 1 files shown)


gibizer (commented):
Do I understand correctly that this fixing a case when the NADs changes but the operator fails to regenerate the network annotations for the deployment? If so the I think we can add an envtest case for this.

stuggi:
Actually I am not sure we need it, the situation I tested was to deploy without the configured NAD to exist. When creating the NAD the controller did not reconcile after the timeout and realized that there is a warning logged that it returns an error and a non-zero result.

The issue for this is that we return the RequeueAfter and an error [1].  When the reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff [2], which can be long. So the RequeueAfter is ignored right now.
We could just change to not return the error in this case and the RequeueAfter is not ignored. (we should check this for other situations as well. seems some operators do that a lot) Or remove the RequeueAfter and add this watch, but that might be to much because in general when there is a change on the NAD, like different IP or some other detail it would not be reflected in the annotation, so a reconcile won't be useful. Manual interaction to delete the deployment would be required in that case, or some further logic to delete the deployment automatically, which I am not sure right now we want to have.

[1] https://github.com/openstack-k8s-operators/placement-operator/blob/main/controllers/placementapi_controller.go#L609
[2] https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler

gibizer:
OK, if the issue is the exponential back-off and not a missing reconcile (just a severely delayed one) then I suggest to drop the error from the return, keep the requeueafter, and log the error before return.

gibizer (approved):
Thanks @stuggi !

PR #262, opened 2024-10-31
Improve automated support for local webhooks

abays:
Reworks `make run-with-webhook` target to trap `SIGINT` and `SIGTERM` so as to call `clean_local_webhook.sh` in such a scenario.  This way, a user can run the operator locally and automatically have the local webhooks removed when they finish running the operator via `ctrl+c`.

Also creates a backup of any existing OLM CSV for the operator while preparing to run the operator locally with webhooks, _if_ the current CSV has a deployment `replicas` count greater than zero and/or there are `webhookdefinitions` present.  That backup can then be easily re-applied to restore OLM CSV defaults via `oc patch` by the user at their convenience.  I left the existing `make` `scale-down-placement-controller-csv` target alone, even though its functionality is covered now in `hack/run_with_local_webhook.sh`.  It seemed that maybe it could still be useful in isolation.

### DIFF
--- a/hack/run_with_local_webhook.sh
+++ b/hack/run_with_local_webhook.sh
@@ -1,6 +1,16 @@
 #!/bin/bash
 set -ex
 
+# Define a cleanup function
+cleanup() {
+    echo "Caught signal, cleaning up local webhooks..."
+    ./hack/clean_local_webhook.sh
+    exit 0
+}
+
+# Set trap to catch SIGINT and SIGTERM
+trap cleanup SIGINT SIGTERM
+
 TMPDIR=${TMPDIR:-"/tmp/k8s-webhook-server/serving-certs"}
 SKIP_CERT=${SKIP_CERT:-false}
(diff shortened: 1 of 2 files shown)


hack/run_with_local_webhook.sh
```diff
@@ -90,3 +100,30 @@ webhooks:
 EOF_CAT
 
 oc apply -n openstack -f ${TMPDIR}/patch_webhook_configurations.yaml
+
+oc apply -n openstack -f ${TMPDIR}/patch_webhook_configurations.yaml
```
mrkisaolamb: this line looks like it's now duplicated

abays: Thanks, I'll fix that

mrkisaolamb: Thanks!

mrkisaolamb (approved):
lgtm

SeanMooney:
so i dont know about you but my workflow was always to rung the operator once with the make target to get everything setup up and then control c it then run it under a debugger.

so automatically cleaning them on contol-c is not the behavior i would generally want.

PR #270, opened 2024-11-26
Added a parameter 'apiTimeout' to allow customization

mrkisaolamb:
to the Apache timeout.

The default is set to 60s which we do set for
HAProxy timeouts currently.

To be able to change the HAProxy value based on the apiTimeout with any update (and not just the first time) the code adds a custom annotation "api.placement.openstack.org/timeout" with the value that was initially set, this way flags it as being set by the placement-operator.

Timeout is global for both api and metadata and they are set from placement lvl

There will be follow up patch in openstack-operator to utilize the method 'SetDefaultRouteAnnotations' to set these default route annotations in openstack-operator

resolve: https://issues.redhat.com/browse/OSPRH-10962

### DIFF
--- a/api/v1beta1/placementapi_webhook.go
+++ b/api/v1beta1/placementapi_webhook.go
@@ -175,3 +175,26 @@ func ValidateDefaultConfigOverwrite(
 	}
 	return errors
 }
+
+// SetDefaultRouteAnnotations sets HAProxy timeout values of the route
+func (spec *PlacementAPISpecCore) SetDefaultRouteAnnotations(annotations map[string]string) {
+	const haProxyAnno = "haproxy.router.openshift.io/timeout"
+	// Use a custom annotation to flag when the operator has set the default HAProxy timeout
+	// With the annotation func determines when to overwrite existing HAProxy timeout with the APITimeout
+	const placementAnno = "api.placement.openstack.org/timeout"
+	valPlacementAPI, okPlacementAPI := annotations[placementAnno]
+	valHAProxy, okHAProxy := annotations[haProxyAnno]
+	// Human operator set the HAProxy timeout manually
+	if !okPlacementAPI && okHAProxy {
+		return
+	}
+	// Human operator modified the HAProxy timeout manually without removing the Placemen flag
+	if okPlacementAPI && okHAProxy && valPlacementAPI != valHAProxy {
+		delete(annotations, placementAnno)
+		placementapilog.Info("Human operator modified the HAProxy timeout manually without removing the Placement flag. Deleting the Placement flag to ensure proper configuration.")
+		return
+	}
+	timeout := fmt.Sprintf("%ds", spec.APITimeout)
+	annotations[placementAnno] = timeout
+	annotations[haProxyAnno] = timeout
+}
--- a/controllers/placementapi_controller.go
+++ b/controllers/placementapi_controller.go
@@ -1265,6 +1265,7 @@ func (r *PlacementAPIReconciler) generateServiceConfigMaps(
 		httpdVhostConfig[endpt.String()] = endptConfig
 	}
 	templateParameters["VHosts"] = httpdVhostConfig
+	templateParameters["TimeOut"] = instance.Spec.APITimeout
 
 	extraTemplates := map[string]string{
 		"placement.conf": "placementapi/config/placement.conf",
--- a/tests/functional/placementapi_controller_test.go
+++ b/tests/functional/placementapi_controller_test.go
@@ -344,6 +344,9 @@ var _ = Describe("PlacementAPI controller", func() {
 			myCnf := cm.Data["my.cnf"]
 			Expect(myCnf).To(
 				ContainSubstring("[client]\nssl=0"))
+			configData := cm.Data["httpd.conf"]
+			Expect(configData).Should(
+				ContainSubstring("TimeOut 60"))
 		})
 
 		It("creates service account, role and rolebindig", func() {
--- a/api/bases/placement.openstack.org_placementapis.yaml
+++ b/api/bases/placement.openstack.org_placementapis.yaml
@@ -48,6 +48,11 @@ spec:
           spec:
             description: PlacementAPISpec defines the desired state of PlacementAPI
             properties:
+              apiTimeout:
+                default: 60
+                description: APITimeout for HAProxy, Apache
+                minimum: 10
+                type: integer
               containerImage:
                 description: PlacementAPI Container Image URL (will be set to environmental
                   default if empty)
--- a/api/v1beta1/placementapi_types.go
+++ b/api/v1beta1/placementapi_types.go
@@ -51,6 +51,12 @@ type PlacementAPISpec struct {
 // PlacementAPISpecCore -
 type PlacementAPISpecCore struct {
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=60
+	// +kubebuilder:validation:Minimum=10
+	// APITimeout for HAProxy, Apache
+	APITimeout int `json:"apiTimeout"`
+
+	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=placement
 	// ServiceUser - optional username used for this service to register in keystone
 	ServiceUser string `json:"serviceUser"`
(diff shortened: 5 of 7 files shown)


controllers/placementapi_controller.go
```diff
@@ -1265,6 +1265,7 @@ func (r *PlacementAPIReconciler) generateServiceConfigMaps(
 		httpdVhostConfig[endpt.String()] = endptConfig
 	}
 	templateParameters["VHosts"] = httpdVhostConfig
+	templateParameters["TimeOut"] = instance.Spec.APITimeout
```
auniyal61, line 1268: I understand this will be eventually be read from `config/crd/bases/placement.openstack.org_placementapis.yaml`, where the default timeout is set.

any reason, we need to maintain same spec at `api/bases/placement.openstack.org_placementapis.yaml`

tests/functional/placementapi_controller_test.go
```diff
@@ -344,6 +344,9 @@ var _ = Describe("PlacementAPI controller", func() {
 			myCnf := cm.Data["my.cnf"]
 			Expect(myCnf).To(
 				ContainSubstring("[client]\nssl=0"))
+			configData := cm.Data["httpd.conf"]
+			Expect(configData).Should(
+				ContainSubstring("TimeOut 60"))
```
auniyal61, line 349: +1

api/v1beta1/placementapi_webhook.go
```diff
@@ -175,3 +175,25 @@ func ValidateDefaultConfigOverwrite(
...
+	}
+	// Human operator modified the HAProxy timeout manually without removing the Placemen flag
+	if okPlacemenAPI && okHAProxy && valPlacementAPI != valHAProxy {
+		delete(annotations, placementAnno)
+		return
+	}
+	timeout := fmt.Sprintf("%ds", spec.APITimeout)
+	annotations[placementAnno] = timeout
+	annotations[haProxyAnno] = timeout
+}
```
gibizer, line 200: This feels sensible. Thanks for taking care of the possible conflicts.

tests/functional/placementapi_controller_test.go
```diff
@@ -344,6 +344,9 @@ var _ = Describe("PlacementAPI controller", func() {
...
 		})
```
gibizer, line 350: Do we have a way to test the annotations on the Route? I guess not in functional env as the Route is only created by the openstack-operator but maybe in the kuttl env?

api/v1beta1/placementapi_webhook.go
gibizer: I'm wondering if the other operators using the same logic. At least they should. And then This function can be abstracted to lib-common as it only need the annotations param and a service specific annotation param but it does not depend on the spec at all.

api/v1beta1/placementapi_webhook.go
```diff
@@ -175,3 +175,25 @@ func ValidateDefaultConfigOverwrite(
...
 	return errors
 }
+
+// SetDefaultRouteAnnotations sets HAProxy timeout values of the route
+func (spec *PlacementAPISpecCore) SetDefaultRouteAnnotations(annotations map[string]string) {
+	const haProxyAnno = "haproxy.router.openshift.io/timeout"
+	// Use a custom annotation to flag when the operator has set the default HAProxy timeout
+	// With the annotation func determines when to overwrite existing HAProxy timeout with the APITimeout
+	const placementAnno = "api.placement.openstack.org/timeout"
+	valPlacementAPI, okPlacemenAPI := annotations[placementAnno]
```
Deydra71: nit: `okPlacemenAPI `should be `okPlacementAPI`

(Noticed it when I started to replicate this for keystone)

api/v1beta1/placementapi_webhook.go
```diff
@@ -175,3 +175,25 @@ func ValidateDefaultConfigOverwrite(
...
+	valHAProxy, okHAProxy := annotations[haProxyAnno]
+	// Human operator set the HAProxy timeout manually
+	if !okPlacemenAPI && okHAProxy {
+		return
+	}
+	// Human operator modified the HAProxy timeout manually without removing the Placemen flag
+	if okPlacemenAPI && okHAProxy && valPlacementAPI != valHAProxy {
+		delete(annotations, placementAnno)
```
bogdando, line 193: would be nice to log the deletion reason

api/v1beta1/placementapi_webhook.go
```diff
@@ -175,3 +175,25 @@ func ValidateDefaultConfigOverwrite(
...
+// SetDefaultRouteAnnotations sets HAProxy timeout values of the route
+func (spec *PlacementAPISpecCore) SetDefaultRouteAnnotations(annotations map[string]string) {
+	const haProxyAnno = "haproxy.router.openshift.io/timeout"
+	// Use a custom annotation to flag when the operator has set the default HAProxy timeout
+	// With the annotation func determines when to overwrite existing HAProxy timeout with the APITimeout
+	const placementAnno = "api.placement.openstack.org/timeout"
+	valPlacementAPI, okPlacemenAPI := annotations[placementAnno]
+	valHAProxy, okHAProxy := annotations[haProxyAnno]
+	// Human operator set the HAProxy timeout manually
+	if !okPlacemenAPI && okHAProxy {
```
bogdando: please log why the annotation wasn't updated

controllers/placementapi_controller.go
bogdando: @gibizer do you know perchance?

api/v1beta1/placementapi_webhook.go
stuggi: iirc, neutron was the first one adding this https://github.com/openstack-k8s-operators/neutron-operator/blob/main/api/v1beta1/neutronapi_webhook.go. we have functional tests as part of the openstack-operaotr  https://github.com/openstack-k8s-operators/openstack-operator/blob/main/tests/functional/ctlplane/openstackoperator_controller_test.go#L855

controllers/placementapi_controller.go
mrkisaolamb: timeout will be set from openstack operator lvl

api/v1beta1/placementapi_webhook.go
mrkisaolamb: > I'm wondering if the other operators using the same logic. At least they should. And then This function can be abstracted to lib-common as it only need the annotations param and a service specific annotation param but it does not depend on the spec at all.

yes all operators use same logic so it can be move to lib-common but maybe as a follow up?

api/v1beta1/placementapi_webhook.go
mrkisaolamb: done

api/v1beta1/placementapi_webhook.go
mrkisaolamb: not sure if this is good idea to log that nothing happend, this sounds like something that make only noise in logs

api/v1beta1/placementapi_webhook.go
mrkisaolamb: thanks done

api/v1beta1/placementapi_webhook.go
gibizer: follow up works for me

controllers/placementapi_controller.go
gibizer: @auniyal61 the commit introduced the coping explains the reason https://github.com/openstack-k8s-operators/nova-operator/pull/450

tests/functional/placementapi_controller_test.go
gibizer: At least we need to extend https://github.com/openstack-k8s-operators/openstack-operator/blob/main/tests/functional/ctlplane/openstackoperator_controller_test.go#L855 when we integrate this patch into the openstack-operator.

gibizer (commented):
I have a testing question still open. Do you have an openstack-operator patch integrating this change into there?

tests/functional/placementapi_controller_test.go
mrkisaolamb: sure, done https://github.com/openstack-k8s-operators/openstack-operator/pull/1220/files

gibizer (approved):
Looks good. cc @dprince as this is an API change (even though a limited one with a single non-nested field)

PR #271, opened 2024-11-26
[make][pre-commit]Check CRD schema to avoid update issues

mrkisaolamb:
The new crd-schema-check make target compares the CRD schema of the patch with the schema on the tip of main and report errors on non backward compatible changes.

This make target now also run in pre-commit both locally and in CI.

This make target uses https://github.com/openshift/crd-schema-checker to do the actual checking.

Related: [OSPRH-11833](https://issues.redhat.com//browse/OSPRH-11833)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -376,3 +376,11 @@ scale-down-placement-controller-csv:
 	if [ "$(has_webhooks)" != "null" ]; then \
 	    oc patch csv -n $(OPERATOR_NAMESPACE) placement-operator.v0.0.1 --type json -p='[{"op": "remove", "path": "/spec/webhookdefinitions"}]'; \
 	fi
+
+CRD_SCHEMA_CHECKER_VERSION ?= release-4.16
+BRANCH=main
+
+PHONY: crd-schema-check
+crd-schema-check: manifests
+	INSTALL_DIR=$(LOCALBIN) CRD_SCHEMA_CHECKER_VERSION=$(CRD_SCHEMA_CHECKER_VERSION) hack/build-crd-schema-checker.sh
+	INSTALL_DIR=$(LOCALBIN) BASE_REF="$${PULL_BASE_SHA:-$(BRANCH)}" hack/crd-schema-checker.sh
--- a/hack/build-crd-schema-checker.sh
+++ b/hack/build-crd-schema-checker.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -euxo pipefail
+
+if [ -f "$INSTALL_DIR/crd-schema-checker" ]; then
+    exit 0
+fi
+
+mkdir -p "$INSTALL_DIR/git-tmp"
+git clone https://github.com/openshift/crd-schema-checker.git \
+    -b "$CRD_SCHEMA_CHECKER_VERSION" "$INSTALL_DIR/git-tmp"
+pushd "$INSTALL_DIR/git-tmp"
+GOWORK=off make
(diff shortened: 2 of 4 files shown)


hack/build-crd-schema-checker.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -euxo pipefail
+
+if [ -f "$INSTALL_DIR/crd-schema-checker" ]; then
+    exit 0
+fi
+
+mkdir -p "$INSTALL_DIR/git-tmp"
+git clone https://github.com/openshift/crd-schema-checker.git \
```
bogdando, line 9: nit: to speed up that you can limit by the max tree depth argunment

gibizer: on my machine with
*  --depth=1 it took 1.9 sec
*  --depth=1 --filter=tree:0 it took 3.8 sec
* without any filtering or limiting it took 1.4 sec

to clone the repo so I think it does not make sense to apply these for this repo. (The repo is probably small enough not to benefit from the extra steps)

mrkisaolamb: @gibizer thanks for checking! Agree this options are reserved for large repos and monorepos especially with big files that are changed at once (eg. some generated big text files that git track as a replacement). Filter can make additional request that will slow down whole cloning process. But interesting is that --depth=1 was slower at least with branch shoudn't be, but maybe it was some network hickup

bogdando:
lgtm

Makefile
```diff
@@ -376,3 +376,10 @@ scale-down-placement-controller-csv:
 	if [ "$(has_webhooks)" != "null" ]; then \
 	    oc patch csv -n $(OPERATOR_NAMESPACE) placement-operator.v0.0.1 --type json -p='[{"op": "remove", "path": "/spec/webhookdefinitions"}]'; \
 	fi
+
+CRD_SCHEMA_CHECKER_VERSION ?= release-4.16
+
+PHONY: crd-schema-check
+crd-schema-check: manifests
+	INSTALL_DIR=$(LOCALBIN) CRD_SCHEMA_CHECKER_VERSION=$(CRD_SCHEMA_CHECKER_VERSION) hack/build-crd-schema-checker.sh
+	INSTALL_DIR=$(LOCALBIN) BASE_REF="$${PULL_BASE_SHA:-$(BRANCH)}" hack/crd-schema-checker.sh
```
gibizer, line 386: you need BRANCH defined in the makefile otherwise the schema diff will not work locally as the base version will be
```
+ git show :config/crd/bases/placement.openstack.org_placementapis.yaml
````
which seems to be valid for git but definitely not returning the main version of the yaml file.

**Review by gibizer — CHANGES_REQUESTED** [review]

File: Makefile (hunk 2 above)
Comment (mrkisaolamb) [reply Makefile:386]: done

File: Makefile [hunk 3]
```diff
@@ -376,3 +376,11 @@ scale-down-placement-controller-csv:
 	if [ "$(has_webhooks)" != "null" ]; then \
 	    oc patch csv -n $(OPERATOR_NAMESPACE) placement-operator.v0.0.1 --type json -p='[{"op": "remove", "path": "/spec/webhookdefinitions"}]'; \
 	fi
+
+CRD_SCHEMA_CHECKER_VERSION ?= release-4.16
+BRANCH=main
```
bogdando, line 381: good to have it for main as of now, but we should also cover cherry picks to fr branches

gibizer (approved):
Thanks. Looks good!

PR #280, opened 2025-02-11
Consume Topology CR by reference

fmount:
This patch provides more granular control over `Pod` placement and scheduling through the `Topology CR` integration introduced in `infra-operator`.
In particular it provides:

- a new `API` parameter (`TopologyRef`) defined for each `Component` that allows to reference an existing `Topology` CR in the same namespace

- the operator logic that retrieves and processes the referenced `Topology CR` through the functions provided by `infra-operator` API

- enhanced  `Deployment` configuration that incorporates the processed `Topology`

- a set of `envTests` to test the lifecycle (add/update/override/remove) of the resulting `Deployment` when a `Topology` is referenced

Note that `webhooks` are in place to prevent referencing a `Topology` from a different namespace (which is not a supported scenario).

Jira: https://issues.redhat.com/browse/OSPRH-13971


mrkisaolamb (commented):
lgtm, but probably this should link only to https://issues.redhat.com/browse/OSPRH-13971

api/v1beta1/placementapi_webhook.go
```diff
@@ -103,12 +115,21 @@ func (r *PlacementAPI) ValidateCreate() (admission.Warnings, error) {
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *PlacementAPI) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
 	placementapilog.Info("validate update", "name", r.Name)
+	errors := field.ErrorList{}
```
gibizer: the empty list set here is unused and overwritten by the line `errors = r.Spec.ValidateUpdate(oldPlacement.Spec, field.NewPath("spec"))`

api/v1beta1/placementapi_webhook.go
```diff
@@ -103,12 +115,21 @@ func (r *PlacementAPI) ValidateCreate() (admission.Warnings, error) {
...
+	basePath := field.NewPath("spec")
+	errors = r.Spec.ValidateUpdate(oldPlacement.Spec, field.NewPath("spec"))
+
+	// When a TopologyRef CR is referenced, fail if a different Namespace is
+	// referenced because is not supported
+	if r.Spec.TopologyRef != nil {
+		if err := topologyv1.ValidateTopologyNamespace(r.Spec.TopologyRef.Namespace, *basePath, r.Namespace); err != nil {
+			errors = append(errors, err)
+		}
+	}
```
gibizer: I think this is in the wrong place. The openstack-operator webhook calls `func (r PlacementAPISpecCore) ValidateUpdate(old PlacementAPISpecCore, basePath *field.Path) field.ErrorList` only the placement-operator webhook would call `func (r *PlacementAPI) ValidateUpdate(old runtime.Object) (admission.Warnings, error) ` but that webhook is not enabled as we rely on the webhooks on OpenStackControlPlane CR only.

https://github.com/openstack-k8s-operators/openstack-operator/blob/f06dc64a8e5eae0918750fc97e22de3989863a90/apis/core/v1beta1/openstackcontrolplane_webhook.go#L416

api/v1beta1/placementapi_webhook.go
```diff
@@ -89,8 +90,19 @@ var _ webhook.Validator = &PlacementAPI{}
...
+
+	errors = r.Spec.ValidateCreate(field.NewPath("spec"))
+
+	// When a TopologyRef CR is referenced, fail if a different Namespace is
+	// referenced because is not supported
+	if r.Spec.TopologyRef != nil {
+		if err := topologyv1.ValidateTopologyNamespace(r.Spec.TopologyRef.Namespace, *basePath, r.Namespace); err != nil {
+			errors = append(errors, err)
+		}
+	}
```
gibizer: this should be in the validate call that is actually called by the openstack-operator webhooks. I suggest to add test coverage in openstack-operator validating that the webhooks works as I think this is not called today in a normal deploymnet.

controllers/placementapi_controller.go
```diff
@@ -1107,8 +1143,45 @@ func (r *PlacementAPIReconciler) ensureDeployment(
...
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.TopologyReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.TopologyReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, fmt.Errorf("waiting for Topology requirements: %w", err)
+	}
+
+	// If TopologyRef is present and ensureNeutronAPITopology returned a valid
```
gibizer: copy pasta

controllers/placementapi_controller.go
```diff
@@ -242,6 +243,11 @@ func (r *PlacementAPIReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 	}()
 
+	// Init Topology condition if there's a reference
+	if instance.Spec.TopologyRef != nil {
+		c := condition.UnknownCondition(condition.TopologyReadyCondition, condition.InitReason, condition.TopologyReadyInitMessage)
+		savedConditions.Set(c)
+	}
```
gibizer: please move the condition initialization to initConditions() where the rest of the conditions are initialized. Also I suggest to avoid the conditional condition initialization and instead always have the condition added to the list and ensureDeployment() should set the condition ready / error depending only on the result of ensurePlacementAPITopology(). This way all the repeated checking of `instance.Spec.TopologyRef != nil' can be encapsulated in ensurePlacementAPITopology

controllers/placementapi_controller.go
```diff
@@ -1107,8 +1143,45 @@ func (r *PlacementAPIReconciler) ensureDeployment(
...
+			condition.SeverityWarning,
+			condition.TopologyReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, fmt.Errorf("waiting for Topology requirements: %w", err)
+	}
+
+	// If TopologyRef is present and ensureNeutronAPITopology returned a valid
+	// topology object, set .Status.LastAppliedTopology to the referenced one
+	// and mark the condition as true
+	if instance.Spec.TopologyRef != nil {
```
gibizer, line 1179: I think we can push all this nil checking down to the ensurePlacementAPITopology

controllers/placementapi_controller.go
```diff
@@ -1107,8 +1143,45 @@ func (r *PlacementAPIReconciler) ensureDeployment(
...
+		// update the Status with the last retrieved Topology name
+		instance.Status.LastAppliedTopology = instance.Spec.TopologyRef.Name
+		// update the TopologyRef associated condition
+		instance.Status.Conditions.MarkTrue(condition.TopologyReadyCondition, condition.TopologyReadyMessage)
+	} else {
+		// remove LastAppliedTopology from the .Status
+		instance.Status.LastAppliedTopology = ""
```
gibizer, line 1186: as the Spec field is a pointer, can we make the Status field symmetrically a pointer as well? Mixing "" and nil is confusing now.

controllers/placementapi_controller.go
```diff
@@ -1318,3 +1391,61 @@ func (r *PlacementAPIReconciler) createHashOfInputHashes(
 	}
 	return hash, changed, nil
 }
+
+// ensurePlacementAPITopology - when a Topology CR is referenced, remove the
+// finalizer from a previous referenced Topology (if any), and retrieve the
+// newly referenced topology object
+func (r *PlacementAPIReconciler) ensurePlacementAPITopology(
```
gibizer, line 1404: Is this in any way depends on the PlacementAPI CR or the PlacementAPIReconciler? As far as I see it isn't. If so can we move this to lib-common or to the topologyv1 module?

pkg/placement/deployment.go
```diff
@@ -163,20 +165,31 @@ func Deployment(
...
+		// Get the Topology .Spec
+		ts := topology.Spec
+		// Process TopologySpreadConstraints if defined in the referenced Topology
+		if ts.TopologySpreadConstraints != nil {
+			deployment.Spec.Template.Spec.TopologySpreadConstraints = *topology.Spec.TopologySpreadConstraints
+		}
+		// Process Affinity if defined in the referenced Topology
+		if ts.Affinity != nil {
+			deployment.Spec.Template.Spec.Affinity = ts.Affinity
+		}
```
gibizer, line 181: This also seems to be generic and not depending on any placement specific data. Can we move this logic to the topology object? E.g. 
```golang
if topology != nil {
  topology.applyTo(deployment)
}

File: tests/functional/placementapi_controller_test.go [hunk 10]
```diff
@@ -878,6 +878,85 @@ var _ = Describe("PlacementAPI controller", func() {
...
+			Eventually(func(g Gomega) {
+				placement := GetPlacementAPI(names.PlacementAPIName)
+				g.Expect(placement.Status.LastAppliedTopology).Should(BeEmpty())
+			}, timeout, interval).Should(Succeed())
+
+			Eventually(func(g Gomega) {
+				g.Expect(th.GetDeployment(names.DeploymentName).Spec.Template.Spec.TopologySpreadConstraints).To(BeNil())
+				g.Expect(th.GetDeployment(names.DeploymentName).Spec.Template.Spec.Affinity).ToNot(BeNil())
+			}, timeout, interval).Should(Succeed())
+		})
```
gibizer, line 1006: Please assert that the new condition is set properly.
Please assert the error cases when topology points to a non existent CR.

controllers/placementapi_controller.go
```diff
@@ -1318,3 +1391,61 @@ func (r *PlacementAPIReconciler) createHashOfInputHashes(
...
+			return nil, err
+		}
+	}
+	// TopologyRef is passed as input, get the Topology object
+	if tpRef != nil {
+		// no Namespace is provided, default to instance.Namespace
+		if tpRef.Namespace == "" {
+			tpRef.Namespace = helper.GetBeforeObject().GetNamespace()
+		}
+		// Build a defaultLabelSelector (service=neutronapi)
```
gibizer: copy pasta

gibizer (changes requested):
I think the webhook change is in the wrong place. And I have a couple of refactoring proposals.

controllers/placementapi_controller.go
fmount: sorry thanks, fixing this

mrkisaolamb:
looks like kuttl job is blocked by this bug: https://issues.redhat.com/browse/RHEL-78362

tests/functional/placementapi_controller_test.go
fmount: ack, I'll do that in my next PS. Thanks, nice suggestion.

pkg/placement/deployment.go
fmount: This is an interesting follow up that I'd like to try it to optimize the code across the board. 
While I'm not sure I'll be able to push all the optimization by FR2 freeze, I'll try hard.
Because I'm spreading the pattern across the operators, I'm wondering if I can take this suggestion as a note and follow up to meet the criteria you're expecting in the operator.

controllers/placementapi_controller.go
fmount: no, this can be seen as an independent function: for that purpose I avoided passing `instance` (because this would be different across operators and across subCRs of the same operator).
Technically, as a follow up, I could optimize the entire code across the operators to move this function to infra-operator. It requires a few more lines (the `defaultSelector` changes based on the operators), but I can follow up and do a small patch series batch to move it to `topologyv1` (we removed `topology` module from lib-common to avoid circular dependencies).
I will remove the dependency on `PlacementAPIReconciler`, because I agree we do not need it, but I might need to follow up with a different PR to remove this function entirely and move it to infra-operator topology API.
As per your other common about common code that is not operator specific, I'm taking a note about this, and I can put a comment if you like, to follow up once I have the feature available to all the operators (that is high priority as per the deadline). I'm planning to land all this first patch series during this week, so I can work on the optimizations you're suggesting next week.

controllers/placementapi_controller.go
fmount: I had to extract it from there because I'm not passing `instance`. I might have to experiment to something that wraps `.Status` to not be operator specific and still be able to move the function in infra-operator as an optimization. Not sure you have a suggestion.

controllers/placementapi_controller.go
fmount: Ok sorry, it makes sense. I'm going to move it under `initConditions`. My first implementation had `TopologyReadyCondition` always defined (with a simpler logic), following the same lifecycle of the other conditions.
I think there was an agreement to not see this at all if no `TopologyRef` is passed, hence the conditional, which is evaluated only when the `.Spec` contains a reference to a `Topology`. 
As I mentioned, I changed the implementation based on other suggestions, but I understand you might not like this approach, and if it is a blocker for placement I can change the implementation (even though it will slightly diverge from the others).

api/v1beta1/placementapi_webhook.go
fmount: oh right, I don't need this. Removing

api/v1beta1/placementapi_webhook.go
fmount: oh! thank you for pointing this out. In my testing I might have seen it working because we have webhooks enabled both at the openstack-operator level and at the service operator level! I'm not sure why but in my mind I was expecting the service operator webhooks being executed, but I have to revisit this check to ensure it flows correctly in both cases.

controllers/placementapi_controller.go
fmount: I can experiment with the change you're requesting. I initially kept it as a string because I reuse `LastAppliedTopology` as `TopoRef.Name` (that can be an empty string) [1], but not a blocker here if it helps to make the code clearer and more readable.

[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/topology/v1beta1/topology_types.go#L133

pkg/placement/deployment.go
olliewalsh: Not just used for deployments so expect it would need to be something like `topology.applyTo(deployment.Spec.Template.Spec)`

fmount: Right, this might be interesting as an update we can rollout to simplify the code.

controllers/placementapi_controller.go
fmount: @gibizer this helps [1] to remove the function for each operator and simplify the logic

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/346

pkg/placement/deployment.go
fmount: This patch [1] should cover this particular aspect.

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/346

controllers/placementapi_controller.go
gibizer: Lets keep the condition conditional for consistency sake. Maybe after the deadlines we can re-discuss if we want to change it to non-conditional everywhere.

api/v1beta1/placementapi_webhook.go
fmount: I will enable the Validations in openstack-operator. The current code takes a `namespace` as input, and it would be tricky to move this function call in `<Service>SpecCore`. 
However, even for testing purposes, I'd like to keep the operator Validate{Create,Update} calls to test them via webhooks when the operator is deployed using envTests. 
Does it make sense to you? Or do you want me to remove webhook calls entirely?

api/v1beta1/placementapi_webhook.go
fmount: right, I'm going to add that coverage in openstack-operator. I was wondering, as per my previous comment, if we want to keep these calls for envTests validations.

controllers/placementapi_controller.go
gibizer: Sorry I missed the fact that TopologyRef is a struct and LastAppliedTopology is just a string. So the real comment form me is that the two fields should have the same type as basically we move the value of TopologyRef to LastAppliedTopology when the reconciler updates the deployment. If we can do that now that would be ideal. But I'm aware of the looming deadline. Unfortunately changing it later is probably a non backward compatible API change. I let you or the podified core decided and I will accept whathever is the outcome.

controllers/placementapi_controller.go
gibizer: Follow up with the refactor sounds good to me as this does not change the behavior of the code. I quickly looked at the pulled out functions and the direction looks good to me.

pkg/placement/deployment.go
gibizer: I agree that this can be an independent follow up after the freeze. And the proposed direction looks good to me.

api/v1beta1/placementapi_webhook.go
gibizer: you don't need to keep these for envtest. In envtest the placement-operator webhooks are running and calling `func (r *PlacementAPI) ValidateCreate() (admission.Warnings, error)` but this function calls `r.Spec.ValidateCreate(field.NewPath("spec"))` where the new `ValidateTopologyNamespace` supposed to be moved. So both the placement-operator local envtest and openstack-operator will work and will reach the same single function call.

api/v1beta1/placementapi_webhook.go
gibizer: We have `r.Spec.ValidateUpdate` called in both cases (envtest, openstack-operator) just from different paths. So if the validation is moved there the it will work in both cases. That function gets a PlacementAPISpecCore (or PlacementAPISpecCore view of a real PlacementAPISpec if called by placement webhook). But as the TopologyRef field is part of PlacementAPISpecCore struct you can always get the TopologyRef.Namespace value. 
Ahh I see now the problem is the availability of r.namespace. But I think that is a generic problem for the case when you want to call this validation from the openstack-operator as that also calls `r.Spec.ValidateUpdate`. So to be able to actually implement the webhook validation for the openstack-operator case you need to extend the current `r.Spec.ValidateUpdate` interface called the openstack-operator to pass down the necessary namespace information anyhow.

controllers/placementapi_controller.go
gibizer: I see now. As I now accept to have the new condition handled conditionally it is less important to hide this from here as well. And I rather keep ensurePlacementAPITopology generic. 

Btw if you are interested how allow passing generic things around then I suggest to look at https://github.com/openstack-k8s-operators/nova-operator/blob/0933c978bad435a91ec3fff1121f762a93e9bba9/controllers/common.go#L184-L187 for an example. You can formulate interfaces and use the interface type as the formal parameter. Then any struct that implements the interface (i.e. has the matching functions) can be passed to the function. I'm not say you need to do this. Just showing a possible way if you want to.

pkg/placement/deployment.go
```diff
@@ -163,20 +165,22 @@ func Deployment(
...
+		// If possible two pods of the same service should not
+		// run on the same worker node. If this is not possible
+		// the get still created on the same worker node.
+		deployment.Spec.Template.Spec.Affinity = affinity.DistributePods(
+			common.AppSelector,
+			[]string{
+				ServiceName,
+			},
+			corev1.LabelHostname,
+		)
```
gibizer, line 192: for the future refactoring PR: If this code is common across operators then this can be also moved to be a centralized default for the case when were there is no Topology object provided.

tests/functional/placementapi_controller_test.go
```diff
@@ -878,6 +878,134 @@ var _ = Describe("PlacementAPI controller", func() {
 		})
 	})
 
+	When("A PlacementAPI is created with topologyref", func() {
```
gibizer: Let use a different name than the test at L916. (Can be fixed in a follow up)

gibizer (changes requested):
* Keeping my -1 due to the validation webhook bug. 
* I let others decide on the Status API change so that is not a blocker for me. Even though it is an unfortunate loose-loose situation
* the rest of my open comments can be fixed in a follow up after the deadline so those are not blocking.

pkg/placement/deployment.go
fmount: Yeah, I was thinking about that but I'm not sure because for a `DaemonSet`, for example, we do not set `PodAntiAffinity` (it gets ignore and it might be the wrong thing to do). For this reason I'm still thinking about it: we could always run `DistributePods` for code consistency (and we know the scheduler ignores it), but for the reader (that describes the DaemonSet resource) might be confusing or simply the wrong thing to do (other than the fact that we do not maintain the k8s scheduler, and we'll don't know how it evolves and if a new addition might generate an issue for us).
@olliewalsh do you have an opinion on this?

controllers/placementapi_controller.go
fmount: So I think I see your point, and as I prepared the infra-operator PR to address the previous question, I can do this as well as a follow up, but I'm not comfortable to apply this change here and diverge from the rest of the operators (I'd like to do some proper testing).
The original idea (and because we'll never support referencing a topology on a different namespaces), was to only keep track of the `name`, and explicitly ignore namespace (especially to not generate false ideas to the user). For this reason `lastAppliedTopology` aligns w/ `TopologyRef.Name`, which is a string (an empty string results in `LastAppliedTopology` being omitted from the `.Status`).
However, I might have time before the code freeze to land a second round of patches, and eventually align to the interface before bumping openstack-operator, so I'll try to do that when I finish this first batch.

controllers/placementapi_controller.go
fmount: Thank you, this is very interesting for something I was thinking (a single storage operator that handles multiple services), and I'll dig more into that. It's a very nice idea and I see you also have a base Reconciler used by the specialized controllers! Thanks for sharing. I think for now https://github.com/openstack-k8s-operators/infra-operator/pull/346 should be sufficient to remove the dedicated `ensurePlacementAPITopology` function (I can /hold this patch until that lands).

mrkisaolamb:
To avoid kuttl fail we need first land this: https://github.com/openstack-k8s-operators/placement-operator/pull/281

api/v1beta1/placementapi_webhook.go
fmount: Yes, I think your summary is correct. To call it properly via openstack-operator we need to extend the current interface and pass `r.Namespace` to the underlying `<Service>SpecCore.Validate<Create|Update>`.
I was thinking about something different to not change the interface.
`ValidateTopology` is a function provided by `infra-operator` topology API, so technically, from openstack-operator, we could do something like:

```golang
t := topologyv1.TopoRef(*r.Spec.Glance.Template.TopologyRef)
errors = append(errors, validateTopologyServiceSpec(&t, basePath.Child("glance").Child("topologyref"), r.Namespace)...)
```
where:

```golang
func validateTopologyServiceSpec(tp *topologyv1.TopoRef, basePath *field.Path, namespace string) field.ErrorList {
	var allErrs field.ErrorList
	// When a TopologyRef CR is referenced, fail if a different Namespace is
	// referenced because is not supported
	if tp != nil {
		if err := topologyv1.ValidateTopologyNamespace(tp.Namespace, *basePath, namespace); err != nil {
			allErrs = append(allErrs, err)
		}
	}
	return allErrs
}
```

The above is an alternative approach to avoid changing the underlying interface, where I need to pass the `namespace`, ... but  actually, adding `...namespace string`) to the ValidaUpdate/ValidateCreate seems a cleaner solution that works both with openstack-operator and in a standalone scenario.
I'll propose this here, and at the same time I'll start a parallel openstack-operator patch, and a series for all the operators where I bump the signature to pass down the namespace. 
I just wanted to share my thoughts regarding the alternative solutions, but I agree passing down the namespace is better.

api/v1beta1/placementapi_webhook.go
fmount: Right, the idea is to move it to move it to the inner level. Not sure I'm following the calls correctly, but from [1] we call the `PlacementAPISpecCore` `ValidateUpdate`, so it should work in both cases (standalone and openstack-operator) if we have the actual code there.

[1] https://github.com/openstack-k8s-operators/openstack-operator/blob/f06dc64a8e5eae0918750fc97e22de3989863a90/apis/core/v1beta1/openstackcontrolplane_webhook.go#L416

tests/functional/placementapi_controller_test.go
fmount: Done

api/v1beta1/placementapi_webhook.go
```diff
@@ -126,33 +127,49 @@ func (r *PlacementAPI) ValidateDelete() (admission.Warnings, error) {
...
 }
 
-func (r PlacementAPISpec) ValidateUpdate(old PlacementAPISpec, basePath *field.Path) field.ErrorList {
-	return r.PlacementAPISpecCore.ValidateCreate(basePath)
+func (r PlacementAPISpec) ValidateUpdate(old PlacementAPISpec, basePath *field.Path, namespace string) field.ErrorList {
+	return r.PlacementAPISpecCore.ValidateCreate(basePath, namespace)
 }
 
-func (r PlacementAPISpecCore) ValidateCreate(basePath *field.Path) field.ErrorList {
+func (r PlacementAPISpecCore) ValidateCreate(basePath *field.Path, namespace string) field.ErrorList {
```
fmount, line 143: @gibizer I tested this update through [1] (I still need to push a single bump once I push the webhooks change for all operators), and I think it works pretty well.

```
# openstackcontrolplanes.core.openstack.org "openstack-galera" was not valid:
# * spec.placement.template.topology[namespace]: Invalid value: "namespace": Customizing namespace field is not supported

```

[1] https://github.com/openstack-k8s-operators/openstack-operator/compare/main...fmount:openstack-operator:topology_webhook#diff-fcb2ae5fc2510a906947042c693c1d054971fa6a901ef6013b4d4c91138a9839R262

pkg/placement/deployment.go
```diff
@@ -163,20 +165,22 @@ func Deployment(
...
-			ServiceName,
-		},
-		corev1.LabelHostname,
-	)
 	if instance.Spec.NodeSelector != nil {
 		deployment.Spec.Template.Spec.NodeSelector = *instance.Spec.NodeSelector
 	}
-
+	if topology != nil {
+		topology.ApplyTo(&deployment.Spec.Template)
```
fmount: @gibizer I tested this code with my last build [1] and I think we can push this kind of change everywhere.
Because my goal is to keep an order in the patches I'm  landing, we have two options here:
1. Land infra-operator first, and update this code to use `topology.ApplyTo( ...`: this means I might need to hold this patch until we land [2]
2.  Revert L172 to the previous version of the code (and restore an `ensurePlacementAPITopology` function), and follow up with a dedicated patch once [2] is merged.

I'm ok either way, just trying to organize the work as the amount of patches to spread updates to all operators (to keep them aligned) requires quite some time.

[1] https://github.com/openstack-k8s-operators/openstack-operator/compare/main...fmount:openstack-operator:topology_webhook#diff-fcb2ae5fc2510a906947042c693c1d054971fa6a901ef6013b4d4c91138a9839R262
[2] https://github.com/openstack-k8s-operators/infra-operator/pull/346

fmount:
@gibizer I think 

> * Keeping my -1 due to the validation webhook bug.
> 
>     * I let others decide on the Status API change so that is not a blocker for me. Even though it is an unfortunate loose-loose situation
> 
>     * the rest of my open comments can be fixed in a follow up after the deadline so those are not blocking.

I think webhook validation is addressed in the last version. Instead of being an unfortunate loose-loose situation I think I might have time to rollout this update https://github.com/openstack-k8s-operators/glance-operator/pull/689 across the board during the week, that pretty much addresses the `.Status.LastAppliedTopology` API change (other fixing the `webhook` check for `.SpecCore` and do other code optimization we already discussed in this PR).

api/v1beta1/placementapi_webhook.go
gibizer: Yeah it looks good now, openstack-operator hits the validation code

controllers/placementapi_controller.go
gibizer: Yeah nova-operator has some common code across reconcilers. It helps a bit with code duplication. Still if I had unlimited time I would go even further and have all the common logic in the operators as generic functions. In the past I did some investigation in my own time how a fully generic and fully modular operator could be built. I think the README here https://github.com/gibizer/okofw summarize the direction of my grand ideas. The reality is that given the current priorities this direction never be really explored further.

>  Thanks for sharing. I think for now openstack-k8s-operators/infra-operator#346 should be sufficient to remove the dedicated ensurePlacementAPITopology function (I can /hold this patch until that lands).

Yes that infra refactor is the sweet spot now to go for. Thanks for doing it.

controllers/placementapi_controller.go
gibizer: OK. I let you decide if this API change fits into the deadline or not.

api/v1beta1/placementapi_webhook.go
gibizer: Yeah the alternative approach would work for openstack-operator but actually force us to call the topology validation from two places from two different ways. I rather extend the interface now, this interface extension might be useful for  future validations as well as namespace is a generic thing

pkg/placement/deployment.go
gibizer: I have no hard preference either. Whatever feels easier for you to land works for me.

gibizer:
> @gibizer I think
> 
> > * Keeping my -1 due to the validation webhook bug.
> >   
> >   * I let others decide on the Status API change so that is not a blocker for me. Even though it is an unfortunate loose-loose situation
> >   * the rest of my open comments can be fixed in a follow up after the deadline so those are not blocking.
> 
> I think webhook validation is addressed in the last version. Instead of being an unfortunate loose-loose situation I think I might have time to rollout this update [openstack-k8s-operators/glance-operator#689](https://github.com/openstack-k8s-operators/glance-operator/pull/689) across the board during the week, that pretty much addresses the `.Status.LastAppliedTopology` API change (other fixing the `webhook` check for `.SpecCore` and do other code optimization we already discussed in this PR).

Yep webhook looks good. If you have time to roll out the API change before the deadline that is good. (I thought we hit the no new feature deadline last Friday). 
Anyhow my last blocking concern is fixed now so this looks good to me. I will double check on slack if we can still land this after the no new feature deadline

gibizer:
/lgtm
So after the slack discussion I think this is good to go.

gibizer (approved):

PR #282, opened 2025-02-13
Improve error handling

mrkisaolamb:
To follow best practices for error and exception handling in Go and to prevent regressions, err113 and gosec checks in golang-ci-linter are now enabled. Errors now use predefined messages for status updates.
Resolve:  https://issues.redhat.com/browse/OSPRH-9063

### DIFF
--- a/controllers/placementapi_controller.go
+++ b/controllers/placementapi_controller.go
@@ -97,7 +97,7 @@ func ensureSecret(
 			return "",
 				ctrl.Result{},
 				*secret,
-				fmt.Errorf("Secret %s not found", secretName)
+				fmt.Errorf("%w: Secret %s not found", err, secretName)
 		}
 		conditionUpdater.Set(condition.FalseCondition(
 			condition.InputReadyCondition,
@@ -113,7 +113,7 @@ func ensureSecret(
 	for _, field := range expectedFields {
 		val, ok := secret.Data[field]
 		if !ok {
-			err := fmt.Errorf("field '%s' not found in secret/%s", field, secretName.Name)
+			err := fmt.Errorf("%w: field '%s' not found in secret/%s", util.ErrFieldNotFound, field, secretName.Name)
 			conditionUpdater.Set(condition.FalseCondition(
 				condition.InputReadyCondition,
 				condition.ErrorReason,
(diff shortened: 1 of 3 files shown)


controllers/placementapi_controller.go
```diff
@@ -66,6 +66,11 @@ import (
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 )
 
+const (
+	//NovaNetworkAttachmentsReadyInitMessage
+	NovaNetworkAttachmentsReadyErrorMessage = "NetworkAttachments error occurred not all pods have interfaces with ips as configured in NetworkAttachments: %s"
```
stuggi: possibly not `Nova` ? see comment bellow, could we just update the one in lib-common?

controllers/placementapi_controller.go
```diff
@@ -1156,13 +1161,12 @@ func (r *PlacementAPIReconciler) ensureDeployment(
...
 	} else {
-		err := fmt.Errorf("not all pods have interfaces with ips as configured in NetworkAttachments: %s", instance.Spec.NetworkAttachments)
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			condition.NetworkAttachmentsReadyCondition,
 			condition.ErrorReason,
 			condition.SeverityWarning,
-			condition.NetworkAttachmentsReadyErrorMessage,
-			err.Error()))
+			NovaNetworkAttachmentsReadyErrorMessage,
+			instance.Spec.NetworkAttachments))
```
stuggi: could we just update https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/condition/conditions.go#L269 ? now that it is only used as the condition error message?

controllers/placementapi_controller.go
mrkisaolamb: done

controllers/placementapi_controller.go
mrkisaolamb: done

mrkisaolamb:
https://github.com/openstack-k8s-operators/openstack-operator/pull/1306  landed, so we should be unblocked on ci

stuggi (approved):
/lgtm

PR #285, opened 2025-02-20
Align LastAppliedTopology to TopologyRef interface

fmount:
While `TopologyRef` is able to reference `Name` and `Namespace`, we currently only keep track of `TopologyRef.Name` in  `.Status.LastAppliedTopology`. This patch aligns both interfaces to use the same type in the same form.
In addition, some logic is moved to infra-operator because is common to all operators, and this patch also adopts this new form.


api/v1beta1/placementapi_types.go
```diff
@@ -232,3 +232,16 @@ func SetupDefaults() {
...
+// processed by the handle topology logic
+func (instance PlacementAPI) GetLastAppliedTopologyRef() *topologyv1.TopoRef {
+	lastAppliedTopologyName := ""
+	if instance.Status.LastAppliedTopology != nil {
+		lastAppliedTopologyName = instance.Status.LastAppliedTopology.Name
+	}
+	return &topologyv1.TopoRef{
+		Name:      lastAppliedTopologyName,
+		Namespace: instance.Namespace,
+	}
```
gibizer, line 246: this feels ove-rcomplicated.
1. the namespace already stored in the instance.Status.LastAppliedTopology so I don't see why we are using the instance.namespace
2. the function needs to return a pointer to the last applied topology, so why don't we just return the field LastAppliedTopology itself? As far as I see EnsureServiceTopology handles lastAppliedTopology being nil

gibizer (changes requested):
One simplification request inline. Otherwise looks OK.

api/v1beta1/placementapi_types.go
fmount: yeah, I didn't remember we modified [1]  to handle `.Status.LastAppliedTopology` as `nil`. I think this code comes from the old `LastAppliedTopology` being a string before moving to the struct.
I'm going to update it to just `instance.GetLastAppliedTopology()`. 
Thank you, good catch.

[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/topology/v1beta1/topology_types.go#L256

fmount: Ah ok now I remember why I did this. I think you're right about removing this logic from the service operators and move it to [1], but because we check `topoRef == nil && lastAppliedTopology != nil` this will segfault by not having an empty topRef. 
I don't think we have the time to update infra-operator, merge it, and land the same update everywhere. The only thing I can propose is to follow up after FR2.
This kind of debt is because we transitioned the `LastAppliedTopology` interface to `topologyv1.TopoRef` without introducing breaking changes in the infra-operator behavior. That `ensureServiceTopology` function needs an update, and this function right now helps to not do that (and do it later, in a follow up).

[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/topology/v1beta1/topology_types.go#L256

gibizer: as this is mostly a refactoring question that does not affect API an external behavior I'm OK to handle this as a follow up.

gibizer (approved):
Looks good. My comment can be handled in a follow up./

api/v1beta1/placementapi_types.go
fmount: FYI https://issues.redhat.com/browse/OSPRH-14489

PR #286, opened 2025-02-27
Switch Topology defaultLabelSelector to lib-common GetLabelSelector

fmount:


gibizer (commented):
Is it a bug fix or just a refactor? If the former then it would be nice to have some test coverage for it.

fmount:
> Is it a bug fix or just a refactor? If the former then it would be nice to have some test coverage for it.

Not a bug, we just realized during a meeting that it would be better, when we pass the defaultLabelSelector, to match the whole set of labels associated to a deployment or a statefulset, to avoid any scheduling risk for other components matching the same label (because it was a single label). I think this simply switches to the new method, but the behavior is unchanged.

mrkisaolamb:
I think we have test coverage here https://github.com/openstack-k8s-operators/placement-operator/blob/main/tests/functional/placementapi_controller_test.go#L499

fmount:
@gibizer @mrkisaolamb are you ok to land this patch? I think I need this update due to the hard timeline, wondering if we can approve it

mrkisaolamb:
I'm ok with this changes because IMO this is tested but lets wait for @gibizer rsponse

bogdando (approved):
lgtm

PR #296, opened 2025-03-26
[run-with-webhook] Handle operator deploy by Openstack initialization

karelyatin:


hack/run_with_local_webhook.sh
```diff
@@ -122,6 +122,17 @@ if [ -n "${CSV_NAME}" ]; then
 
     oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 0}]"
     oc patch "${CSV_NAME}" -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/webhookdefinitions', 'value': []}]"
+else
```
SeanMooney, line 125: oh someoen finally got a round to figuring out what we need to change to fix this :)

i wonder if we really need to keep the fallback to the old way of doing it or if we could just update.

perhaps we should add a todo to remove the old way eventually?

we do not have EUS support so we should not need to deploy any of the older release going forward so we could likely clean this up but i don't object to keeping both for the time being.

gibizer: Yeah and on FR1 branch we have the old way preserved anyhow.

bogdando:
Thank you, this is LGTM (aside of the comment to replace the 'main' stanza with the 'else' section)

hack/run_with_local_webhook.sh
karelyatin: The old way still needed if standalone operators are deployed i.e make nova, make neutron etc instead of make openstack and make openstack_init

gibizer (approved):

PR #297, opened 2025-03-27
add basic zuul jobs

SeanMooney:
This change adds 2 new zuul jobs
first we leaverge the common zuul kuttl job that is designed
to run kuttl form in tree. to enable that i have ported
the webhook install fix form nova-operator

second the nova-operator-tempest-multinode job
is used as a base for a placemnet version to provide tempest
coverage on placement operator changes.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -1,4 +1,62 @@
 ---
+- job:
+    name: placement-operator-kuttl
+    parent: cifmw-multinode-kuttl-operator-target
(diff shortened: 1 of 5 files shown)


SeanMooney:
i think we need to first add the noop job and then this to make speculative execution work so ill do that now in a separate pr.

gibizer:
placement installer code is moved, please propose this to nova-operator repo if it is still needed.

PR #298, opened 2025-03-27
add zuul noop jobs

SeanMooney:
add safe jobs we know can't fail so that we can use
zuuls speculitve merge feature to add complex ones
later.


gibizer (approved):
Safe first step

PR #303, opened 2025-03-31
Improve topology testing coverage

fmount:
This patch improves our topology testing with several enhancements and extends the `topologyRef` tests:
   - `finalizer` verification for each sub component that references a topology
   - Updated `Topologies` to support `component-specific` label selectors

This work completes the topology validation improvements initiated in [1], addressing the remaining open points.

Jira: https://issues.redhat.com/browse/OSPRH-15220

[1] nova-operator#941

### DIFF
--- a/tests/functional/base_test.go
+++ b/tests/functional/base_test.go
@@ -24,9 +24,11 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	placementv1 "github.com/openstack-k8s-operators/placement-operator/api/v1beta1"
 	"github.com/openstack-k8s-operators/placement-operator/pkg/placement"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type Names struct {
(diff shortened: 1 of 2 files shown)


bogdando:
LGTM

tests/functional/base_test.go
```diff
@@ -195,17 +197,35 @@ func GetSampleTopologySpec() map[string]interface{} {
...
+				},
+			},
+		},
+	}
+	return topologySpec, topologySpecObj
 }
 
 // CreateTopology - Creates a Topology CR based on the spec passed as input
-func CreateTopology(topology types.NamespacedName, spec map[string]interface{}) client.Object {
+func CreateTopology(
```
gibizer, line 225: Hm, both CreateTopology and GetTopology seems to be copy paste across the service operators. So it is a good candidate to move it to a common helper in infra-operator https://github.com/openstack-k8s-operators/infra-operator/tree/main/apis/test/helpers

gibizer (approved):
Looks good. Please as a separate step move the common test helper to infra-operator

tests/functional/base_test.go
fmount: Correct. I agree with you on this and I have it tracked in https://issues.redhat.com/browse/OSPRH-14532, not only for Topology, but in general for service operators, so we have an interface we can use for envTests. GetTopology is a good example of that.
Unfortunately, because I need to balance the capacity among multiple parallel things, that Jira is still in my backlog, and some copy/paste is still required.

PR #315, opened 2025-04-28
Use ReaderFailOnMissingInformer for cache options

bogdando:
When controller makes queries with client.{Get,List} on resources haven’t been declared upfront, controller-runtime will initialize an informer on-the-fly and block on warming up its cache. This leads to issues like:

* controller-runtime starting a watch for a resource type and start caching all its objects in memory (even if you were trying to query only one resource), potentially leading to the process running out of memory.
* unpredictable reconciliation times while the informer cache is syncing, during which your worker goroutine will be blocked from reconciling other resources.

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,6 +32,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
+	cache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -97,6 +98,7 @@ func main() {
 			BindAddress: metricsAddr,
 		},
 		HealthProbeBindAddress: probeAddr,
+		Cache:                  cache.Options{ReaderFailOnMissingInformer: true},
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "73d6b7ce.openstack.org",
 		WebhookServer: webhook.NewServer(


bogdando:
This is related to recommendations made here https://ahmet.im/blog/controller-pitfalls/#understand-the-cached-clients
Let's see if this works. If it fails with missing informers, when we should adjust [ctrl.NewControllerManagedBy](https://github.com/bogdando/placement-operator/blob/a3a7cfc496096c487a30aaf52ea95550dee9fc52/controllers/placementapi_controller.go#L906)
to start building always static caches upfront.
We could extend this to other operators as well.
@stuggi @gibizer @mrkisaolamb @dprince @bshephar wdyt?

stuggi:
With this change all the operators would have to watch e.g. KeystoneAPI (which they currently don't), right?
```
        +      Kind=KeystoneAPI is not cached
        +    reason: Error
        +    severity: Warning
        +    status: "False"
             type: ServiceConfigReady
```
I don't think they strictly have to. They do the get to validate that a keystoneAPI exists and get the endpoints from there, but as soon they exist, they are static and would not have to watch and reconcile on changes.

mrkisaolamb:
I'm not sure if we will see the gains mentioned in the blog post. We are only checking Keystone and MariaDB resources, and even if a resource is not immediately available, we simply reconcile again. Once these resources are created, they will only be deleted when we delete the Placement resource — we don't expect them to be removed otherwise.

Also, if we really want to disable caching completely, we would need to make more changes, because clearly something else is happening — the kuttl test failures seem to be directly related to these changes.

BTW Thanks Bogdan to bring this up, it's definitely good to keep this behavior in mind.

bogdando:
> I'm not sure if we will see the gains mentioned in the blog post. We are only checking Keystone and MariaDB resources, and even if a resource is not immediately available, we simply reconcile again. Once these resources are created, they will only be deleted when we delete the Placement resource — we don't expect them to be removed otherwise.
> 
> Also, if we really want to disable caching completely, we would need to make more changes, because clearly something else is happening — the kuttl test failures seem to be directly related to these changes.
> 
> BTW Thanks Bogdan to bring this up, it's definitely good to keep this behavior in mind.

No, I don't think we should disable caching completely

bogdando:
> With this change all the operators would have to watch e.g. KeystoneAPI (which they currently don't), right?
> 
> ```
>         +      Kind=KeystoneAPI is not cached
>         +    reason: Error
>         +    severity: Warning
>         +    status: "False"
>              type: ServiceConfigReady
> ```
> 
> I don't think they strictly have to. They do the get to validate that a keystoneAPI exists and get the endpoints from there, but as soon they exist, they are static and would not have to watch and reconcile on changes.

We don't have a lot of KeystoneAPI to watch, so I don't see this potential change as a problem. On the other hand, by following the rule of it is better to be explicit than implicit, this provides a guardrail against future additions of list()/get() operations without updating the "allow list" first

PR #320, opened 2025-05-14
Do not update the status when reconciler exits due to panic

ratailor:
As of now, we use deffered call to always update the status field when Reconcile call exist, which seems correct for any other error except when Reconciler exits due to panic.

This change adds a call to recover function and try to handle panic and log error.

Closes: [OSPRH-12982](https://issues.redhat.com//browse/OSPRH-12982)


bogdando (commented):
LGTM as it is fully aligned with that https://github.com/openstack-k8s-operators/nova-operator/pull/956 contains

gibizer (approved):
Looks good

PR #364, opened 2025-11-11
Application Credential support

Deydra71:
Jira: [OSPRH-16629](https://issues.redhat.com/browse/OSPRH-16629)

Adds the end-to-end support for consuming Keystone ApplicationCredentials (AC) in the Placement operator, enabling Placement pods to use AC-based authentication when available.

API changes:

Adds an optional authentication field to the Placement CR:

`spec.auth.applicationCredentialSecret` — name of the Secret that contains the Keystone Application Credential ID and Secret (AC_ID and AC_SECRET).

Reconcile behavior:

Reads `spec.auth.applicationCredentialSecret`
Attempts to load AC_ID / AC_SECRET from the referenced Secret (via the Keystone helper).
If the secret is missing or incomplete, it falls back to password authentication (the AppCred auth is optional, not an error).
- Once the AC Secret is ready with valid AC_ID and AC_SECRET fields, templates AC credentials into Placement service configuration
- Computes hash of Secret contents and stores in configVars to trigger rolling updates when credentials rotate

Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/567


Makefile
```diff
@@ -54,7 +54,7 @@ OPERATOR_SDK_VERSION ?= v1.41.1
 DEFAULT_IMG ?= quay.io/openstack-k8s-operators/placement-operator:latest
 IMG ?= $(DEFAULT_IMG)
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.31
```
mrkisaolamb, line 57: why we downgrade this requirements?

test/functional/placementapi_controller_test.go
```diff
@@ -1438,4 +1438,46 @@ var _ = Describe("PlacementAPI reconfiguration", func() {
...
+			)
+			DeferCleanup(th.DeleteInstance, acSecret)
+		})
+
+		It("should render ApplicationCredential auth in placement.conf", func() {
+			configSecret := th.GetSecret(names.ConfigMapName)
+			conf := configSecret.Data["placement.conf"]
+
+			Expect(conf).To(ContainSubstring("application_credential_id = test-ac-id"))
+			Expect(conf).To(ContainSubstring("application_credential_secret = test-ac-secret"))
```
mrkisaolamb, line 1517: please add also negative assert for username, password etc, also to be sure that we cover all changes in conf we should also assert auth_type

mrkisaolamb (commented):
Other than these small comments, everything looks good. So once we land the keystone-operator changes, we should be ready to go with this patch as well. I don't see any issues with the update/upgrade path. The only thing worth adding might be a kuttl test with the new secret

Makefile
Deydra71: Thanks for catching this! Probably an old remnant of local testing..

templates/placementapi/config/placement.conf
```diff
@@ -15,14 +15,20 @@ connection = {{ .DatabaseConnection }}
...
-project_name = service
-username = {{ .ServiceUser }}
-password = {{ .PlacementPassword }}
 www_authenticate_uri = {{ .KeystonePublicURL }}
 auth_url = {{ .KeystoneInternalURL }}
+{{ if .UseApplicationCredentials -}}
+auth_type = v3applicationcredential
+application_credential_id = {{ .ACID }}
+application_credential_secret = {{ .ACSecret }}
+{{ else -}}
```
auniyal61, line 24: +1, changes are reflecting.

auniyal61 (commented):
the kuttl test fail - `GetACSecretName` exist in keystone operator now
its because of dependency bump, can be in next PS
along with Kamil comment on test

Rest of the PR lgtm and works in my local deployment, thanks

mrkisaolamb (approved):
lgtm

auniyal61:
recheck

mrkisaolamb (approved):
lgtm

PR #371, opened 2025-11-25
Add pprof-bind-address flag support

stuggi:
Add optional pprof profiling endpoint configuration via --pprof-bind-address flag, matching neutron-operator implementation. Disabled by default (empty string).


abays (approved):
/lgtm

PR #376, opened 2025-12-19
Bump golangci-lint to 2.7.2 for placement

dprince:
## Summary
-Bump golangci-lint to 2.7.2

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,7 +38,7 @@ repos:
     - id: go-mod-tidy
 
 - repo: https://github.com/golangci/golangci-lint
-  rev: v2.4.0
+  rev: v2.7.2
   hooks:
     - id: golangci-lint-full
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -38,7 +38,7 @@ repos:
     - id: go-mod-tidy
 
 - repo: https://github.com/golangci/golangci-lint
-  rev: v2.4.0
+  rev: v2.7.2
```
stuggi, line 41: should also bump https://github.com/openstack-k8s-operators/placement-operator/blob/main/.pre-commit-config.yaml#L41 ?

dprince: That is this line right :)

stuggi: right :)

stuggi (approved):
/lgtm

PR #392, opened 2026-02-10
Allow triggering the build workflow manually of needed

stuggi:
There was an issue in github which is probably the root cause why an image for a merged PR was not running. This PR allows to trigger the job from the Actions tab in github.


abays (approved):
/lgtm

PR #393, opened 2026-02-12
Sync envtest optimization across all operators

stuggi:
With recent changes to the CI env we hit issues running envtest hitting timeouts in test since 25s was not enough, but also during kube-api startup and register custom CRDs before checks start and teardown of the kube-api. Therefore adding ControlPlaneStartTimeout and ControlPlaneStopTimeout bumping the default 60s to 120s.

Also hitting slow env and with our tests, we have seen that the default cluster ip range /24 is to small, increasing to /12

Also disable not required admission plugins for envtest: ResourceQuota: This stops the "quota evaluation timed out" error. The API server will no longer try to calculate how many resources exist before letting you create new ones.

ServiceAccount: This prevents the API server from looking for a ServiceAccount issuer or secret for every pod/resource. Since envtest doesn't run the ServiceAccount controller, this plugin often just causes unnecessary errors or delays.

NamespaceLifecycle: (Optional but recommended) This allows you to create resources in a namespace even if the namespace is in a "Terminating" state. This is very helpful when your cleanup is slow due to CPU throttling.

### DIFF
--- a/test/functional/suite_test.go
+++ b/test/functional/suite_test.go
@@ -114,6 +114,10 @@ var _ = BeforeSuite(func() {
 
 	By("bootstrapping test environment")
 	testEnv = &envtest.Environment{
+		// Increase this to 60 or 120 seconds for the single-core run
+		ControlPlaneStartTimeout: 120 * time.Second,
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

PR #411, opened 2026-04-21
Use OpenStack CR operatorOverrides to scale down placement operator

karelyatin:
Switch from CSV-based scaling to using OpenStack CR operatorOverrides
  feature for scaling down the placement operator. This avoids the need
  to disable the openstack init operator and provides better integration
  with the OpenStack operator management system.


SeanMooney:
oh i do not have an env to test this myself currently but this makes sense
i was going to say we are not allwo to have a depency liek this but this is actully just a chagne toe the run_with_webhook integration for local deployment not the operator its self so that is completely fine in the make/bash script we use here.

it would have nice to call that out a little more explcity in the PR body

im not going to appove this mearly be cause we do not have ci testign fo this as far as im aware and i have not verified it works locally but i have no objection to this change.

amartyasinha:
Hi @karelyatin!
placement-operator repo no longer accepts changes to main branch, as its code is now maintained in nova-operator repo. Please open a PR in nova-operator repo for these changes and close this PR.

gibizer:
placement installer code is moved, please propose this to nova-operator repo.

PR #420, opened 2026-06-02
Disable automatic dependency bump renovate and gh workflow

amartyasinha:


.github/workflows/force-bump-pr-scheduled.yaml
```diff
@@ -1,15 +0,0 @@
-name: Scheduled Force Bump PR
-
-on:
-  schedule:
-    - cron: '0 6 * * 6'  # 6AM UTC Saturday
-
-jobs:
-  call-build-workflow:
-    if: github.ref == 'refs/heads/main' && github.repository_owner == 'openstack-k8s-operators'
-    uses: openstack-k8s-operators/openstack-k8s-operators-ci/.github/workflows/force-bump-branches.yaml@main
```
gibizer, line 10: https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/blob/9e9bb98a2878ac7477c6bbf788aa2b0cf2511de4/.github/workflows/force-bump-branches.yaml#L20

This is also used to trigger bump on fr5. We need to keep that. Until the last point release is released form fr5. As far as I understand FR6 is 18.0.21 so the last FR5 based bugfix release is 18.0.20. It is not released yet but it is expected to be released soon on 06.10. To be safe lets wait for 18.0.20 release and then drop this.

Also this means, after 18.0.20 FR5 branch will become unused. And placement-operator should not get an FR6 branch. So at that we can stop bumping on the FR5 branch as well

gibizer (commented):
Lets wait until next week's last FR5 based bugfix release.

.github/workflows/force-bump-pr-scheduled.yaml
SeanMooney: ya i think that makes sense
we might want to give a head up to release delivery or whomever would normally create that branch.

amartyasinha: Thanks @gibizer @SeanMooney! I'll wait until the last release from FR5 is made, and then we can come back to this PR. Maybe I can also add the drop of gh workflow `release-branch-sync` which syncs the `olive` branch.

gibizer: yes as soon as we know that no more releases from FR5 branch (and we already know that no more releases from main) we can drop the olive sync.

amartyasinha: @gibizer Are we good to be back at reviewing this PR? @stuggi recently removed `release-branch-sync` gh workflow, so I think we should be good to remove other gh workflows through this PR.

gibizer: 18.0.20. is cancelled so nothing is expected form -fr5 branch any more. Lets kill this with fire. :)

gibizer (approved):

gibizer:
I forcemerge it, we don't need to pass CI on this repo any more.

PR #423, opened 2026-06-08
Delete Release Branch Sync workflow

stuggi:
Starting with FR5 we no long rely on the specific release sync branch. Lets remove the job.

Jira: [OSPRH-25747](https://redhat.atlassian.net/browse/OSPRH-25747)


gibizer (approved):

stuggi:
will override the override the kuttl job since there are no longer placement targets in the install_yamls main branch Makefile `make: *** No rule to make target 'placement_kuttl'.  Stop.`

